PR #1, opened 2024-01-22
Initial CDK package

zekehuntergreen:
## What does this change?
Adds the beginnings of a cdk package into monorepo


marjisound (approved):
Nice work

PR #3, opened 2024-01-25
Add Github Actions Workflow to publish a whisper docker container to github packages

philmcmahon:
## What does this change?
From my testing so far, for some reason whisper seems to load a bit faster when run from a docker container. So far I have tried:
 - Checking out whisper.cpp on a machine, compiling then running a transcription using jfk.wav and the medium model - slow load times (over a minute)
 - `docker pull` a container generated by this action with jfk.wav and medium model - fast load time (under 10 seconds)

There is probably something we can tweak in the way we are 'installing' whisper.cpp but I think there might be some benefit to using a docker container regardless - as it allows us to have the code that determines what models are available here in our repo rather than buried in [amigo](https://github.com/guardian/amigo).

I am hoping that this work combined with https://github.com/guardian/amigo/pull/1397 will help get us started running whisper on EC2 (note to anyone out of the team who ends up reading this: we found running on equivalently sized EC2 instances much faster than using Fargate)

The GHA workflow in this PR is almost entirely copy pasted from https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages - just with the 'platform' setting added

## How to test
I've tested that the github action build works and tried running the container on a test instance

You can see the packages output by this role here https://github.com/orgs/guardian/packages?repo_name=transcription-service

### DIFF
--- a/.github/workflows/build-whisper-docker.yml
+++ b/.github/workflows/build-whisper-docker.yml
@@ -5,9 +5,10 @@ name: Build and publish whisper docker image
 on:
   workflow_dispatch:
   push:
-    branches: ['whisper-docker', 'main']
+    branches: ['main']
     paths:
       - "whisper_container/Dockerfile"
+      - ".github/workflows/build-whisper-docker.yml"
 
 # Defines two custom environment variables for the workflow. These are used for the Container registry domain, and a name for the Docker image that this workflow builds.
 env:
--- a/.github/workflows/build-whisper-docker.yml
+++ b/.github/workflows/build-whisper-docker.yml
@@ -3,8 +3,11 @@ name: Build and publish whisper docker image
 
 # Configures this workflow to run every time a change is pushed to the branch called `release`.
 on:
+  workflow_dispatch:
   push:
-    branches: ['whisper-docker']
+    branches: ['whisper-docker', 'main']
+    paths:
+      - "whisper_container/Dockerfile"
 
 # Defines two custom environment variables for the workflow. These are used for the Container registry domain, and a name for the Docker image that this workflow builds.
 env:
@@ -19,7 +22,7 @@ jobs:
     permissions:
       contents: read
       packages: write
-      #
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -46,7 +49,8 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: whisper_container/
-          platforms: linux/amd64,linux/arm64
+          # to add x86: linux/amd64
+          platforms: linux/arm64
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/build-whisper-docker.yml
+++ b/.github/workflows/build-whisper-docker.yml
@@ -40,9 +40,10 @@ jobs:
       # It uses the `context` parameter to define the build's context as the set of files located in the specified path. For more information, see "[Usage](https://github.com/docker/build-push-action#usage)" in the README of the `docker/build-push-action` repository.
       # It uses the `tags` and `labels` parameters to tag and label the image with the output from the "meta" step.
       - name: Build and push Docker image
-        uses: docker/build-push-action@f2a1d5e99d037542a71f64918e516c093c6f3fc4
+        uses: docker/build-push-action@v5
         with:
           context: whisper_container/
+          platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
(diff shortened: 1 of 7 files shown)


.github/workflows/build-whisper-docker.yml
```diff
@@ -0,0 +1,57 @@
...
+  workflow_dispatch:
+  push:
+    branches: ['main']
+    paths:
+      - "whisper_container/Dockerfile"
+      - ".github/workflows/build-whisper-docker.yml"
+
+# Defines two custom environment variables for the workflow. These are used for the Container registry domain, and a name for the Docker image that this workflow builds.
+env:
+  REGISTRY: ghcr.io
```
itsibitzi, line 15: Might want to think about using AWS's image repository since you're copying the entire model into the container - the bandwidth between ECR and your eventual EC2 machine should be much better than to ghcr.

philmcmahon: ah great shout, will do

philmcmahon: Though actually with https://github.com/guardian/amigo/pull/1397 the impact of this latency will be reduced as we'll be 'baking' the container into the AMI

itsibitzi: Doesn't this mean you require an amigo build every time you release a new version? Do we have the infra to coordinate that? Wouldn't it be easier to pull the docker image when you start the image so you don't need to extend the CI pipeline? The EC2 machine will need to pull the data either way, either as an XXL sized AMI or as a docker image.

itsibitzi: Sorry, I don't actually have the data - do we see higher bandwidth when pulling AMIs than in the final runtime EC2 VM? 

I guess it would be possible that they throttle down the network controller once your machine is up and running based on the instance type - and downloading the AMI is done full tilt.

itsibitzi: If AMIs are mounted as a network device and AWS has them available "hot" (e.g. the hypervisor can just connect to an AMI with no further loading) then this would indeed be a lot faster. 

It's surprisingly difficult to find out the internals of how AMIs are loaded onto an EC2 instance.

whisper_container/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu:22.04
+WORKDIR /opt
+LABEL com.theguardian.transcription-service.whisper-container="Whisper.cpp container with models downloaded, including ffmpeg"
+
+RUN apt-get update
+RUN apt-get install -y ffmpeg curl wget git build-essential
```
philmcmahon, line 6: remove curl?

philmcmahon:
I've updated this PR to publish to both ECR and GHCR for now. I thought it might be nice to keep publishing to GHCR so that on the off chance anyone wants to try run this outside the guardian the containers are available. In practice it's a pretty simple dockerfile and we will end up [paying](https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages) for data transfer costs for the whisper models so I'm not sure we want to keep doing that going forwards.

.github/workflows/build-whisper-docker.yml
```diff
@@ -0,0 +1,73 @@
...
+      # It uses the `tags` and `labels` parameters to tag and label the image with the output from the "meta" step.
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v5
+        with:
+          context: whisper_container/
+          # to add x86: linux/amd64
+          platforms: linux/arm64
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
marjisound, line 71: Is this step pushing to github & AWS? I'm just trying to get my head around where the push to Amazon ECR is happening 🤔

packages/cdk/lib/repository.ts
```diff
@@ -0,0 +1,97 @@
...
+					],
+				}),
+			},
+		});
+
+		new CfnOutput(this, 'AccessRoleArn', {
+			value: repoAccessRole.roleArn,
+		});
+	}
+}
```
marjisound, line 97: We were looking with @zekehuntergreen and couldn't fully understand how this stack is uploaded to riffraff S3 bucket. We couldn't find it in the build? 🤔

.github/workflows/build-whisper-docker.yml
philmcmahon: I've added a comment to better explain this

packages/cdk/lib/repository.ts
philmcmahon: Just added a new riffraff upload step to deploy the repository. Thanks for this suggestion!

philmcmahon:
I think this might be ready for a final review!

marjisound (approved):
Great work 👏

PR #4, opened 2024-01-29
Add transcription worker autoscaling group and associated typescript app

philmcmahon:
## What does this change?
This PR adds the initial infrastructure for a pool of workers to run whisper.cpp. It has two main components:

## Autoscaling group
The ASG is set up to use spot instances in order to reduce running costs, which makes the CDK quite a bit more complex than for a normal ASG. The main new parts are the launch template - which would normally be generated for us automatically by a GuAutoscalingGroup - and the `mixedInstancesPolicy` - which includes settings related to using spot instances. Unfortunately the GuCDK project does not currently support autoscaling groups with a mixedInstancesPolicy, so I am using the basic CDK ASG. 

I've given worker instances 50GB of disk space - this is a hard limit on the size of files we will be able to transcribe unless we find a way to do some kind of streaming based transcription. 

The main concern with using spot instances is that, should demand for our chosen instance type increase, our instance may be terminated by AWS. To reduce this risk, rather than specifying a single instance type we specify a list of preferred instance types and the autoscaling group will decide which type to use based off how much capacity there is of that type available. We ideally want our transcription to run on c7g.4xlarge instances, but I suspect that performance will be similar on c6g.4xlarge or m7g.4xlarge instances (yet to be tested) - so have included those in the pool. I've also included .8xlarge instances on the off chance that their spot price is so low that it is within our $0.6202/hr budget (which is the on-demand price for a c7g.4xlarge instance)

We can control whether spot instances are actually used via the `onDemandPercentageAboveBaseCapacity` parameter - currently it is set to 100 which means we will only use on demand instances, but the intention is to have it set to 0 once we have enough failure management bits in place to tolerate an instance being terminated part way through a transcription.

## Worker typescript app
This is a very basic app just to test the deployment pipeline (CI>RiffRaff>UserData) for the workers. Unfortunately as we are using a basic AutoscalingGroup rather than a GuAutoscalingGroup the CDK cannot generate the riff-raff.yaml automatically so I have checked in a new riff-raff.yaml file which is a copy of the auto generated one, with an extra s3-upload step to upload the worker app. 

I've tested that the download/run step works. Ultimately we'll probably want to modify the Userdata so that the node process gets run by a user other than root, but this will do for now. 

## How to test
I've tested this by deploying to CODE. For the PROD deploy a manual cloudformation update step will be required to add the AMI parameter.  Once https://github.com/guardian/amigo/pull/1397 is deployed I'll make a new transcription service AMIgo recipe to use


package.json
```diff
@@ -7,6 +7,8 @@
 		"format": "prettier --write 'packages/**/*.ts'",
 		"api::build": "npm run build --workspace api; npm run build-frontend --workspace api",
 		"api::package": "npm run package --workspace api",
+		"worker::build": "npm run build --workspace worker; npm run build --workspace api",
```
zekehuntergreen, line 10:
```suggestion
		"worker::build": "npm run build --workspace worker; npm run build --workspace worker",
```

packages/cdk/lib/transcription-service.ts
```diff
@@ -1,22 +1,32 @@
-import { GuApiLambda } from '@guardian/cdk';
```
zekehuntergreen, line 1: might need to run `npm run format` on this file

packages/cdk/lib/transcription-service.ts
```diff
@@ -67,5 +77,86 @@ export class TranscriptionService extends GuStack {
...
+			InstanceType.of(InstanceClass.M7G, InstanceSize.XLARGE4),
+			InstanceType.of(InstanceClass.C7G, InstanceSize.XLARGE8),
+			InstanceType.of(InstanceClass.C6G, InstanceSize.XLARGE8)
+		]:  [
+			InstanceType.of(InstanceClass.T4G, InstanceSize.MEDIUM)
+		]
+
+
+		// unfortunately GuAutoscalingGroup doesn't support having a mixedInstancesPolicy so using the basic ASG here
+		new AutoScalingGroup(this, "TransciptionWorkerASG", {
```
zekehuntergreen, line 129:
```suggestion
		new AutoScalingGroup(this, "TranscriptionWorkerASG", {
```

packages/cdk/lib/transcription-service.ts
```diff
@@ -67,5 +77,86 @@ export class TranscriptionService extends GuStack {
...
+				availabilityZones: ["eu-west-1a", "eu-west-1b", "eu-west-1c"],
+			}),
+			vpcSubnets: {
+				subnets: GuVpc.subnetsFromParameter(this, {
+					type: SubnetType.PRIVATE,
+					app: workerApp
+				})
+			},
+			// instances should shut themselves down when they have finished the transcription - they should not be terminated
+			// by the ASG
```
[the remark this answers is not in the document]
philmcmahon: ah yes this comment is a bit out of date - I wrote it before our discussion. I've updated this comment to what I think is more accurate - I _think_ it still makes sense to automatically turn on scale in protection until the instance has had a chance to start the 'worker' app which can then decide whether there's any work to do or not, but don't have a strong feeling about this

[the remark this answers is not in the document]
zekehuntergreen: yeah I agree we should start with scale-in protection on, then turn it on / off depending on whether or not a transcription is happening

.github/workflows/api-ci.yaml
```diff
@@ -25,6 +25,10 @@ jobs:
         run: |
           npm run api::build
           npm run api::package
+      - name: Package Worker
+        run: |
+          npm run worker::build
+          npm run worker::package
       - name: CDK synth
```
marjisound, line 32: Do you think it'd be better if rename this file now that it contains the infra for both api & worker?

packages/cdk/lib/transcription-service.ts
```diff
@@ -67,5 +92,92 @@ export class TranscriptionService extends GuStack {
...
+				]
+			: [InstanceType.of(InstanceClass.T4G, InstanceSize.MEDIUM)];
+
+		// unfortunately GuAutoscalingGroup doesn't support having a mixedInstancesPolicy so using the basic ASG here
+		new AutoScalingGroup(this, 'TranscriptionWorkerASG', {
+			minCapacity: 0,
+			maxCapacity: isProd ? 20 : 4,
+			autoScalingGroupName: `transcription-service-workers-${this.stage}`,
+			vpc: GuVpc.fromIdParameter(this, 'InvestigationsInternetEnabledVpc', {
+				availabilityZones: ['eu-west-1a', 'eu-west-1b', 'eu-west-1c'],
```
marjisound, line 154: How does this know what VPC to choose from? Is there supposed to be a parameter in the stack named `InvestigationsInternetEnabledVpc`?

packages/worker/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "worker",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "esbuild --bundle --platform=node --target=node20 --outfile=dist/index.js src/index.ts",
+    "package": "cd dist; mkdir -p ../target; zip -qr ../target/worker.zip ./*"
+  },
```
marjisound, line 9: It might be good to add a .gitignore under worker so ignore target contents. For api we created a seperate .gitignore too

marjisound (approved):
Looks amazing, all good 🥳  just left 2 comments which is a question and a small suggestion

packages/cdk/lib/transcription-service.ts
philmcmahon: This is a bit of GuCDK magic - behind the scenes it looks for a parameter called /account/primary/vpc/id in parameter store. `InvestigationsInternetEnabledVpc` is just a cdk id here - it is not used for anything

marjisound: Thanks for explaining 👍

PR #11, opened 2024-02-01
Add SQS queue and target-tracking scaling policy

zekehuntergreen:
## What does this change?
- Adds an SQS queue to the stack which can be written to by the API lambda and read from by the worker ASG
- Adds bash script and config files to create and update a target-tracking scaling policy following [this approach](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-target-tracking-metric-math.html). This policy will make the autoscaling group scale in and out so that the number of instances at any given time matches the number of transcription jobs in the queue. Unfortunately this feature hasn't been added to cloudformation or CDK yet so we need to create the policy manually. 

## How to test
tested on CODE infra by 
- creating sqs messages `aws sqs send-message --queue-url [...]transcription-service-task-queue-CODE.fifo --message-body 'this is a test'`
- watching ASG scale out
- reading, then deleting messages
  - `aws sqs receive-message --queue-url [...]transcription-service-task-queue-CODE.fifo`
  - `aws sqs delete-message --queue-url [...]transcription-service-task-queue-CODE.fifo --receipt-handle [value]`
- ASG scales in (after scale-in protection removed from instances)

### DIFF
--- a/packages/cdk/target_tracking_scaling_policy/config-PROD.json
+++ b/packages/cdk/target_tracking_scaling_policy/config-PROD.json
@@ -1,0 +1,67 @@
+{
+	"CustomizedMetricSpecification": {
+		"Metrics": [
+			{
+				"Label": "Number of visible messages",
+				"Id": "m1",
+				"MetricStat": {
+					"Metric": {
+						"MetricName": "ApproximateNumberOfMessagesVisible",
+						"Namespace": "AWS/SQS",
+						"Dimensions": [
+							{
+								"Name": "QueueName",
+								"Value": "transcription-service-task-queue-PROD.fifo"
+							}
+						]
+					},
+					"Stat": "Sum"
+				},
+				"ReturnData": false
+			},
+			{
+				"Label": "Number of invisible (in-flight) messages",
+				"Id": "m2",
+				"MetricStat": {
+					"Metric": {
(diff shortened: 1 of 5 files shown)


packages/cdk/target_tracking_scaling_policy/config-PROD.json
```diff
@@ -0,0 +1,67 @@
+{
```
philmcmahon, line 1: could you avoid having the duplicated configuration by having a generic -STAGE file which the put-scaling-policy script copies, uses sed to set the correct stage, then runs the cli command and deletes?

packages/cdk/target_tracking_scaling_policy/put-scaling-policy.sh
```diff
@@ -0,0 +1,15 @@
+STAGE=$1
```
philmcmahon, line 1: best to add a shebang line here

philmcmahon (approved):
target tracking

philmcmahon:
Great work on this!

packages/cdk/target_tracking_scaling_policy/config-PROD.json
zekehuntergreen: good idea!

PR #13, opened 2024-02-01
Add SQS read/write functionality to API and worker apps

philmcmahon:
This pr adds:
 - A new common library to faciliate sharing code (sqs, types and config code so far) between the api and worker packages
 - A new docker-compose file to spin up localstack locally so that we can run the app locally without having to create DEV queues in AWS
 - A new 'send-message' API endpoint to post (a currently hard coded dummy message) to the SQS queue
 - A new Dockerfile with [fpm](https://fpm.readthedocs.io/) installed on it - this. combined with the `.fpm` config file is used in our worker `package` step to construct a .deb file made up of index.js and a systemd service file. (Using a service file allows us to run the worker process as the ubuntu user rather than root, to restart it on failure (not sure we need this but might be helpful), and to more easily integrate with https://github.com/guardian/devx-logs)
 - Code in the worker app to fetch a message, wait 3 minutes and then delete the message from the queue, to simulate a (rather fast) transcription

## How to test
I've tested this locally and on CODE


package.json
```diff
@@ -5,10 +5,12 @@
 	"main": "index.js",
 	"scripts": {
 		"format": "prettier --write 'packages/**/*.ts'",
+		"api::build": "npm run build --workspace api; npm run build-frontend --workspace api",
```
marjisound, line 8: build-frontend can be removed from this script. Probably it came through after the merge conflict resolving

packages/common/src/sqs.ts
```diff
@@ -0,0 +1,158 @@
...
+			error,
+			errorMsg,
+		};
+	}
+};
+
+export const deleteMessage = async (
+	client: SQSClient,
+	queueUrl: string,
+	reciptHandle: string,
```
marjisound, line 130: type? reciptHandle => receiptHandle ?

marjisound (approved):
Looks great. Just left 2 minor comments 

giphy

PR #14, opened 2024-02-02
Tell riffraff which encrypted AMI to use

philmcmahon:
## What does this change?
AMIgo bakes 3 encrypted AMIs for our account so we need to tell riffraff which one to use


marjisound (approved):
Nice one

PR #17, opened 2024-02-05
Add initial s3 infrastructure, zod types

philmcmahon:
## What does this change?
Paired with @marjisound and @zekehuntergreen 

This PR:
 - Adds a new S3 bucket to hold source media for the transcription service
 - Adds permissions to the worker to fetch and delete from the bucket
 - Updates packages/common/src/types.ts to include all relevant fields we think will be needed to be passed between our different services

## How to test
This code doesn't actually do anything yet so I think is safe to merge if the build passes. I've deployed to CODE to check that the deploy works

### DIFF
--- a/packages/cdk/lib/transcription-service.ts
+++ b/packages/cdk/lib/transcription-service.ts
@@ -10,10 +10,10 @@ import { GuCname } from '@guardian/cdk/lib/constructs/dns';
 import { GuVpc, SubnetType } from '@guardian/cdk/lib/constructs/ec2';
 import {
 	GuAllowPolicy,
-	GuGetS3ObjectsPolicy,
 	GuInstanceRole,
 	GuPolicy,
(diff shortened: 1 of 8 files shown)


marjisound (approved):

packages/cdk/lib/transcription-service.ts
```diff
@@ -56,6 +61,34 @@ export class TranscriptionService extends GuStack {
...
+		});
+
+		// we only want one dev bucket so only create on CODE
+		if (props.stage === 'CODE') {
+			const sourceMediaBucketDev = new GuS3Bucket(
+				this,
+				'TranscriptionServiceUploadsBucket',
+				{
+					app: APP_NAME,
+					bucketName: `transcription-service-source-media-DEV`,
```
zekehuntergreen, line 84:
```suggestion
					bucketName: `transcription-service-source-media-dev`,
```

PR #18, opened 2024-02-05
Fix various bugs in queue code

philmcmahon:
## What does this change?
This change:
 - Enables [persistence](https://docs.localstack.cloud/user-guide/state-management/persistence/) for localstack so we don't have to keep recreating the queue every time the localstack container is restarted
 - Adjusts the timeout/delete timings for SQS with the intention of reducing the risk of messages ending up 'invisible' in the queue for long periods when running locally. 
 - Fixes the queue name in the purge script

## How to test
Tested locally

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,16 +1,17 @@
-version: "3.8"
+version: '3.8'
 
 services:
   localstack:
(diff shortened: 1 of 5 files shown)


marjisound (approved):

philmcmahon:
Update - this PR now also removes 2 unwanted package-lock files

PR #20, opened 2024-02-05
Add initial SNS CDK, localstack configuration and test message 

philmcmahon:
## What does this change?
This PR:
 - Adds CDK to create an SNS topic
 - Removes s3 from our localstack container (following advice from @twrichards that localstack S3 can be challenging to work with due to differences in API implementations) -  note we weren't actually using localstack for S3 yet anyway
 - Adds SNS to localstack, updates setup.sh script to create a topic
 - Adds a new sns.ts client file in the worker project
 - Adds some code to send a dummy event to the SNS topic on app startup

## How to test
I've tested this by deploying to AWS and then starting a worker and verifying that a message gets sent to SNS

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,7 @@ services:
     environment:
       # LocalStack configuration: https://docs.localstack.cloud/references/configuration/
       - DEBUG=${DEBUG:-0}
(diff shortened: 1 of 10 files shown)


marjisound (approved):
Nice work 🥳

PR #21, opened 2024-02-06
Allow client to upload files to S3

zekehuntergreen:
## What does this change?
Adds ability for a user to upload a file they want to be transcribed from the client directly to an S3 bucket.
- new api endpoint `/signedUrl` which takes a `fileName` query parameter, creates a presigned Url, and returns the url in its response
- update to cdk to give media upload buckets CORS policies allowing upload from `transcribe.[stage.dev-]gutools.co.uk`
- update to cdk to enable [transfer acceleration](https://aws.amazon.com/blogs/compute/uploading-large-objects-to-amazon-s3-using-multipart-upload-and-transfer-acceleration/) on media upload buckets. This should speed up uploads especially from browsers far away Ireland where the bucket is hosted.
- update to home page (behind auth) adding a form with a single file input

Also includes a few refactors 
- rename the common package `backend-common`. This will contain code shared between backend services (lambdas and workers). Introduce new `common` package for code that will be shared across all other packages, including `client`.
- override Express's type `Request.User` to include `email` property. This lets us avoid asserting that req.user is `any` in api

To do (in another PR): [multipart upload to s3](https://aws.amazon.com/blogs/compute/uploading-large-objects-to-amazon-s3-using-multipart-upload-and-transfer-acceleration/)

## How to test
run api and client locally, select a file and submit the form

## Images
Screenshot 2024-02-07 at 15 40 44


packages/api/src/index.ts
```diff
@@ -57,6 +67,34 @@ const getApp = async () => {
...
+			const key = uuidv4();
+			const queryParams = SignedUrlQueryParams.safeParse(req.query);
+			if (!queryParams.success) {
+				res.status(422).send('missing query parameters');
+				return;
+			}
+
+			const presignedS3Url = await getSignedUrl(
+				s3Client,
+				new PutObjectCommand({
```
philmcmahon, line 82: This is a nitpick but as a pattern I think it's good to keep code using the aws sdk within the relevant file (in this case `s3.ts`. Here you could pass Bucket/Key/Metadata into getSignedUrl and generate the PutObjectCommand there.

The only actual reason I can think for doing this is that if in some alternative universe where we write some tests, it will be easier to test the `getSignedUrl` function rather than having to test the entire `/signedUrl` controller

packages/api/src/types/express/index.d.ts
```diff
@@ -0,0 +1,13 @@
...
+export {};
+
+// eslint-disable-next-line @typescript-eslint/no-namespace
+declare global {
+	namespace Express {
+		interface User {
+			email: string;
+		}
+	}
+}
```
philmcmahon, line 13: 👍🏻

packages/backend-common/src/s3.ts
```diff
@@ -1,7 +1,8 @@
 import { S3Client } from '@aws-sdk/client-s3';
 
-export const getClient = (region: string) => {
+export const getS3Client = (region: string) => {
 	return new S3Client({
 		region,
+		useAccelerateEndpoint: true,
```
philmcmahon, line 6: ⏩ 🌏

packages/client/src/components/UploadForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const uploadToS3 = async (url: string, blob: Blob) => {
+	try {
+		const response = await fetch(url, {
+			method: 'PUT',
+			body: blob,
+		});
+		const status = await response.status;
+		console.log('upload success:', status);
+		return status == 200;
```
philmcmahon, line 14: ===

packages/client/src/components/UploadForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { useState } from 'react';
+import { SignedUrlResponseBody } from '@guardian/transcription-service-common';
+
+const uploadToS3 = async (url: string, blob: Blob) => {
+	try {
+		const response = await fetch(url, {
+			method: 'PUT',
+			body: blob,
+		});
+		const status = await response.status;
```
philmcmahon, line 12: I _think_ that response.status should just be an integer not a promise https://developer.mozilla.org/en-US/docs/Web/API/Response/status

packages/client/src/components/UploadForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+		console.log('upload success:', status);
```
philmcmahon, line 13: We don't have any client side logging capture stuff so probably not worth logging the happy path.

packages/client/src/components/UploadForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+		return status == 200;
+	} catch (error) {
+		console.error('upload error:', error);
+		return false;
+	}
+};
+
+export const UploadForm = ({ auth }: { auth: AuthState }) => {
+	console.log(auth);
```
philmcmahon, line 22: remove

packages/client/src/components/UploadForm.tsx
zekehuntergreen: oops - good call

packages/api/src/index.ts
zekehuntergreen: good idea - I've moved to s3.ts

src/main.ts
philmcmahon, line 1: ah nice cleanup here

philmcmahon (approved):
Nice!

uploading...

packages/cdk/lib/transcription-service.ts
```diff
@@ -70,6 +71,13 @@ export class TranscriptionService extends GuStack {
 			{
 				app: APP_NAME,
 				bucketName: `transcription-service-source-media-${this.stage.toLowerCase()}`,
+				cors: [
+					{
+						allowedOrigins: [`https://${domainName}`],
```
marjisound, line 76: Does this not affect the worker getObhect call?

packages/client/src/app/page.tsx
```diff
@@ -1,10 +1,15 @@
 'use client';
 import React from 'react';
+import { UploadForm } from '../components/UploadForm';
+import { useContext } from 'react';
+import { AuthContext } from './template';
 
 const Home = () => {
+	const auth = useContext(AuthContext);
```
marjisound, line 8: Nice 👍

packages/client/src/components/UploadForm.tsx
```diff
@@ -0,0 +1,75 @@
...
+		event.preventDefault();
+
+		const maybeFileInput = document.querySelector(
+			'input[name=file]',
+		) as HTMLInputElement;
+		if (!maybeFileInput) {
+			return;
+		}
+		const files = maybeFileInput.files;
+		if (files == undefined || files.length == 0 || !files[0]) {
```
marjisound, line 33: files.length === 0 ?

marjisound (approved):
Looks good, just left 2 minor questions

packages/cdk/lib/transcription-service.ts
zekehuntergreen: I don't think it will. CORS rules control which origin urls browsers can access the bucket from (and allowed headers, methods, etc.).
I don't think requests from other backend services should be affected.

PR #22, opened 2024-02-07
Add transcription to worker

marjisound:
paired with: @philmcmahon 
## What does this change?

This PR does the following:
- Gets file from S3 based on the S3 key that was sent through by the SQS message
- Creates a docker container based on the docker image `ghcr.io/guardian/transcription-service` 
- Converts the file to wav using ffmpeg within the container
- Gets the duration of the file from ffmpeg result
- Updates the `VisibilityTimeout` of the SQS message based on the duration of the file
- Transcribes a file using whisper.cpp within the container (sets the number of threads to use during computation)
- Adds the transcription result to the message that is published to SNS topic
- Set message visibility timeout to 0 if the worker fails to process the job

**TODO in upcoming PR:**
- The scaling policy rule is not working properly, so when a message is added to the queue, the ASG is not automatically triggered. This will be fixed in an upcoming PR
- The worker needs to add scale-in protection to the instance when starting to process 
- The worker needs to remove scale-in protection when it finished the work or when it fails

## How to test

Tested locally and in PROD
- Send a message through api endpoint 
- increase the desired instance of the ASG from 0 to 1
- wait to receive the notification or ssh into the instance and check the logs by `journalctl -u transcription-service-worker`

### DIFF
--- a/packages/cdk/lib/transcription-service.ts
+++ b/packages/cdk/lib/transcription-service.ts
@@ -195,9 +195,6 @@ export class TranscriptionService extends GuStack {
 		// basic placeholder commands
 		userData.addCommands(
 			[
-				// this should be done in AMIgo but for now need to do linux postinstall here https://docs.docker.com/engine/install/linux-postinstall/
-				`usermod -aG docker ubuntu`,
-				`newgrp docker`,
 				`export STAGE=${props.stage}`,
 				`export AWS_REGION=${props.env.region}`,
 				`aws s3 cp s3://${GuDistributionBucketParameter.getInstance(this).valueAsString}/${props.stack}/${props.stage}/${workerApp}/transcription-service-worker_1.0.0_all.deb .`,
--- a/packages/cdk/lib/transcription-service.ts
+++ b/packages/cdk/lib/transcription-service.ts
@@ -195,6 +195,9 @@ export class TranscriptionService extends GuStack {
 		// basic placeholder commands
 		userData.addCommands(
 			[
+				// this should be done in AMIgo but for now need to do linux postinstall here https://docs.docker.com/engine/install/linux-postinstall/
+				`usermod -aG docker ubuntu`,
+				`newgrp docker`,
 				`export STAGE=${props.stage}`,
 				`export AWS_REGION=${props.env.region}`,
 				`aws s3 cp s3://${GuDistributionBucketParameter.getInstance(this).valueAsString}/${props.stack}/${props.stage}/${workerApp}/transcription-service-worker_1.0.0_all.deb .`,
--- a/packages/worker/src/index.ts
+++ b/packages/worker/src/index.ts
@@ -23,6 +23,8 @@ const main = async () => {
 	try {
 		const config = await getConfig();
 
+		const numberOfThreads = config.app.stage === 'PROD' ? 16 : 2;
+
 		const client = getSQSClient(
 			config.aws.region,
 			config.aws.localstackEndpoint,
@@ -81,6 +83,7 @@ const main = async () => {
 			containerId,
 			ffmpegResult.wavPath,
 			fileToTranscribe,
+			numberOfThreads,
 		);
 
 		const transcriptionOutput: TranscriptionOutput = {
--- a/packages/worker/src/index.ts
+++ b/packages/worker/src/index.ts
@@ -14,64 +14,55 @@ import { convertAndTranscribe } from './transcribe';
 const main = async () => {
 	const config = await getConfig();
 
-	const s3Client = getS3Client(config.aws.region);
-	const fileToTranscribe = await getFile(
-		s3Client,
-		`transcription-service-source-media-${config.app.stage.toLowerCase()}`,
-		'tifsample.wav',
-		config.app.stage === 'DEV' ? `${__dirname}/sample` : '/tmp',
-	);
-
-	await convertAndTranscribe(fileToTranscribe);
-
 	const client = getSQSClient(config.aws.region, config.aws.localstackEndpoint);
-
-	// to simulate a transcription job, delay 5 seconds in DEV, 2 minutes in PROD before deleting the message
-	const dummyDelay = config.app.stage === 'DEV' ? 5 : 120;
-	// override timeout to allow enough time for our dummy delay before message becomes available
-	const timeoutOverride = dummyDelay + 1;
-	const message = await getNextMessage(
-		client,
-		config.app.taskQueueUrl,
-		timeoutOverride,
-	);
-
+	const message = await getNextMessage(client, config.app.taskQueueUrl);
 	const snsClient = getSNSClient(
 		config.aws.region,
 		config.aws.localstackEndpoint,
 	);
-
-	if (isFailure(message)) {
+	if (isFailure(message) || !message.message) {
+		console.log('Failed to fetch message or no messages available');
 		return;
 	}
-	console.log(config.app.taskQueueUrl);
+
+	const job = parseTranscriptJobMessage(message.message);
+	console.log(
+		`Fetched transcription job with id ${message.message.MessageId}}`,
+		job,
+	);
+	if (!job) {
+		console.error('Failed to parse message', message);
+		return;
+	}
+	const s3Client = getS3Client(config.aws.region);
+
+	const fileToTranscribe = await getFile(
+		s3Client,
+		config.app.sourceMediaBucket,
+		job?.s3Key,
+		config.app.stage === 'DEV' ? `${__dirname}/sample` : '/tmp',
+	);
+	const text = await convertAndTranscribe(fileToTranscribe);
+	console.log(text);
 
 	await publishTranscriptionOutput(
 		snsClient,
 		config.app.destinationTopicArns.transcriptionService,
 		{
 			id: 'test-id',
-			transcriptionSrt: 'test-srt',
+			transcriptionSrt: text,
 			languageCode: 'en',
 			userEmail: 'test@test.com',
 		},
 	);
 
 	if (message.message?.ReceiptHandle) {
-		const job = parseTranscriptJobMessage(message.message);
-		console.log(
-			`Fetched transcription job with id ${message.message.MessageId}}`,
-			job,
+		console.log(`Deleting message ${message.message?.MessageId}`);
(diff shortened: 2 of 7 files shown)


packages/common/src/s3.ts
```diff
@@ -1,7 +1,33 @@
...
+) => {
+	try {
+		const destinationPath = `${workingDirectory}/${path.basename(key)}`;
+		const data = await client.send(
+			new GetObjectCommand({
+				Bucket: bucket,
+				Key: key,
+			}),
+		);
+		(data.Body as Readable).pipe(createWriteStream(destinationPath));
```
philmcmahon, line 26: I think you might need to wrap this in a promise otherwise it will return instantly before the file has finished writing. You should be able to check this by downloading a big file and making sure that it hangs here rather than immediately logging 'successfully retrieved file from S3 into '

packages/common/src/sqs.ts
```diff
@@ -93,6 +94,29 @@ export const sendMessage = async (
...
+	});
+
+	try {
+		await client.send(command);
+		console.log(
+			`Successfully updated the VisibilityTimeout of the message to ${timeoutOverride}`,
+		);
+	} catch (error) {
+		const errorMsg = 'Failed to update VisibilityTimeout for message';
+		console.error(errorMsg, error);
```
philmcmahon, line 116: is it worth throwing the error here? I think if we fail to update the visibility timeout we should abandon the transcription rather than carrying on and risk it being done by another machine

packages/worker/src/index.ts
```diff
@@ -5,60 +5,124 @@ import {
...
 				client,
 				config.app.taskQueueUrl,
-				message.message?.ReceiptHandle as string,
+				message.message.ReceiptHandle,
 			);
-		}, dummyDelay * 1000);
+		}
+	} catch (error) {
+		const msg = 'Worker failed to complete';
+		console.error(msg, error);
```
philmcmahon, line 111: not for this PR but I think eventually we will want to update a custom 'FailedTranscriptions' metric at this point which we can alarm on

packages/worker/src/index.ts
```diff
@@ -5,60 +5,124 @@ import {
...
 	}
 };
 
+const getFileFromS3 = async (config: TranscriptionConfig, s3Key: string) => {
+	const s3Client = getS3Client(config.aws.region);
+
+	const fileToTranscribe = await getFile(
```
philmcmahon, line 118: tiny thing but at the moment this is a fairly generic function so you could just call it `file` rather than `file ToTranscribe`

packages/worker/src/transcribe.ts
```diff
@@ -0,0 +1,190 @@
...
+	return newContainer.stdout.trim();
+};
+
+export const convertToWav = async (
+	containerId: string,
+	file: string,
+): Promise<FfmpegResult> => {
+	const fileName = path.basename(file);
+	const filePath = `${CONTAINER_FOLDER}/${fileName}`;
+	const wavPath = `${CONTAINER_FOLDER}/output.wav`;
```
philmcmahon, line 90: maybe we should do ${fileName}-converted.wav rather than output.wav in case the same machine does more than one file so we can tell them apart

packages/worker/src/transcribe.ts
```diff
@@ -0,0 +1,190 @@
...
+	console.log(`containerId: ${containerId}`);
+	console.log('file path: ', filePath);
+	console.log('wav file path: ', wavPath);
+
+	try {
+		console.log('calling ffmpeg');
```
philmcmahon, line 96: remove

packages/worker/src/transcribe.ts
```diff
@@ -0,0 +1,190 @@
...
+			duration,
+		};
+	} catch (error) {
+		console.log('ffmpeg failed error:', error);
+		throw error;
+	}
+};
+
+const getDuration = (ffmpegOutput: string) => {
+	const reg = /Duration: (\d{1,2}):(\d{1,2}):(\d{1,2}).\d{1,2},/.exec(
```
philmcmahon, line 126: Nice regex 🥳

packages/worker/src/index.ts
```diff
@@ -5,60 +5,124 @@ import {
...
+
+		if (isFailure(message)) {
+			console.error(`Failed to fetch message due to ${message.errorMsg}`);
+			return;
+		}
+
+		if (!message.message) {
+			console.log('No messages available');
+			return;
+		}
```
philmcmahon, line 45: much better to separate these! nice.

packages/cdk/lib/transcription-service.ts
```diff
@@ -166,6 +166,9 @@ export class TranscriptionService extends GuStack {
 		// basic placeholder commands
 		userData.addCommands(
 			[
+				// this should be done in AMIgo but for now need to do linux postinstall here https://docs.docker.com/engine/install/linux-postinstall/
+				`usermod -aG docker ubuntu`,
+				`newgrp docker`,
```
philmcmahon, line 171: Now that https://github.com/guardian/amigo/pull/1406 is in we can remove this change

packages/worker/src/index.ts
marjisound: Created a ticket for this and added to discuss in planning column

packages/common/src/s3.ts
zekehuntergreen: Could we parse with zod to avoid type assertion?

marjisound: Good spot 👍

packages/worker/src/index.ts
```diff
@@ -5,60 +5,126 @@ import {
...
 		);
-		// wait for 3 minutes then delete the message to simulate the transcription
-		setTimeout(() => {
+
+		if (!job) {
+			console.error('Failed to parse job message', message);
+			return;
+		}
+
+		const fileToTranscribe = await getFileFromS3(config, job?.s3Key);
```
zekehuntergreen, line 59: at this point job isn't can't be undefined, right?
```suggestion
		const fileToTranscribe = await getFileFromS3(config, job.s3Key);
```

packages/worker/src/transcribe.ts
```diff
@@ -0,0 +1,189 @@
...
+				console.error(
+					`failed with code ${result.code} due to: ${result.stderr}`,
+				);
+				reject(result);
+			}
+		});
+	});
+};
+
+export const createContainer = async (tempDir: string): Promise<string> => {
```
zekehuntergreen, line 58: non-blocking: might be clearer to rename `getOrCreateContainer`

packages/worker/src/index.ts
```diff
@@ -5,60 +5,126 @@ import {
...
+			ffmpegResult.duration &&
+			ffmpegResult.duration !== 0
+		) {
+			// Adding 300 seconds (5 minutes) and the file duration
+			// to allow time to load the whisper model
+			await changeMessageVisibility(
+				client,
+				config.app.taskQueueUrl,
+				message.message.ReceiptHandle,
+				ffmpegResult.duration + 300,
```
zekehuntergreen, line 79: should we consider adding a higher overhead than 5 minutes?
maybe something like `ffmpegResult.duration * 1.5 + 300` to reduce the risk that a long-running transcription results in another worker picking up the job?

packages/worker/src/index.ts
zekehuntergreen: should we also mark the sqs message as visible again?

packages/worker/src/index.ts
philmcmahon: I'm not sure what the right figure here is but in theory, if running on our target instance type and if our limited performance testing is accurate for other file types, we should be transcribing files in less than half the length of the file, so I think `duration + 300` might be a good first guess.

Having said that, there's no real reason to abandon a file just because it's taking 2x our expected time so maybe we should set this to a v high figure, I guess it depends what's more likely - a transcription taking longer than expected vs some error happening that causes the transcript to hang for ages at which point it might be better to allocate it to another worker. I suspect probably the first option is more likely! e.g. we haven't tested transcribe time for files in languages other than english

zekehuntergreen: ok, I thought our average transcription times were about the same as the length of the files rather than half!
happy to keep the current visibility timeout and we can adjust while testing if necessary

packages/worker/src/index.ts
```diff
@@ -5,60 +5,126 @@ import {
...
-	// to simulate a transcription job, delay 5 seconds in DEV, 2 minutes in PROD before deleting the message
-	const dummyDelay = config.app.stage === 'DEV' ? 5 : 120;
-	// override timeout to allow enough time for our dummy delay before message becomes available
-	const timeoutOverride = dummyDelay + 1;
-	const message = await getNextMessage(
-		client,
-		config.app.taskQueueUrl,
-		timeoutOverride,
-	);
+	try {
```
zekehuntergreen, line 25: non-blocking: maybe as we test it might be worth moving to doing more specific error handling with try / catches around each step (get a message, retrieve file, convert to wav, etc.)
We might want to handle errors differently based on the specific exception and which step we're on (e.g. if there's a network error while fetching the file we could retry, but if while converting to wav we find the audio file is malformed, we might want to give up and send a failure notification)

marjisound: Yes definetely, currently every step has its own try catch but it's just logging the error and re-throwing it, so that it'll get caught by the outermost try/catch. When we do more testing and recognise that in some of the steps we need to do differently when an error happens, we can then add actions other than just logging.

packages/worker/src/index.ts
marjisound: right, made the change 👍

packages/worker/src/index.ts
marjisound: Thanks both for giving your ideas. For now I leave to as `duration + 300` but will with a bigger not-english file today and see if it takes any longer

packages/common/src/s3.ts
marjisound: Good idea. Made a change for this, does it look OK?

zekehuntergreen: thanks! looks great

zekehuntergreen (approved):
looks good!

obama

PR #28, opened 2024-02-08
Use tiny model to transcribe on CODE

philmcmahon:
## What does this change?
To make testing transcription a bit quicker, this sets us up to use the `tiny` whisper model on CODE. As we are using less powerful workers on CODE we need to use a smaller model so that transcription doesn't take _forever_

As part of this I've also tried to speed up the builds for the docker container by using github actions cache - https://docs.docker.com/build/cache/backends/gha/

## How to test
Deploy to CODE, run a transcription. Leaving this as a draft pending the main worker transcription PR being merged

### DIFF
--- a/.github/workflows/build-whisper-docker.yml
+++ b/.github/workflows/build-whisper-docker.yml
@@ -6,8 +6,8 @@ on:
   workflow_dispatch:
   push:
     paths:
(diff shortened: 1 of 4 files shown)


marjisound (approved):
Thanks for doing it, it'll speed up our testing in code 👍

PR #29, opened 2024-02-12
Add/remove scale in protection for worker instances

marjisound:
## What does this change?

This PR adds scale-in polict to the worker instances and removes that when it's not needed any more. Changes are as followed:
- add a function that can update the scale-in protection of the running instance
- Updates the ASG in cdk to create the instances with scale-in protection 
- Adds new allow policy to the ASG instance role in order to allow to SetInstanceProtection from within the instance
- Removes scale-in protection when  reading a message from queue fails or when there is no message in queue
- Removes scale-in protection when the worker finishes the work
- Removes scale-in protection when the worker fails to finish the work

## How to test

This was tested locally and in PROD

### DIFF
--- a/packages/worker/src/asg.ts
+++ b/packages/worker/src/asg.ts
@@ -2,7 +2,7 @@ import {
 	AutoScalingClient,
 	SetInstanceProtectionCommand,
 } from '@aws-sdk/client-auto-scaling';
-import { readFile } from './transcribe';
+import { readFile } from './util';
 
 export const updateScaleInProtection = async (
 	region: string,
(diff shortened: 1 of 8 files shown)


packages/worker/src/asg.ts
```diff
@@ -0,0 +1,33 @@
...
+			console.log(`instanceId: ${instanceId}`);
+			const autoScalingClient = new AutoScalingClient(clientConfig);
+			const input = {
+				InstanceIds: [instanceId.trim()],
+				AutoScalingGroupName: `transcription-service-workers-${stage}`,
+				ProtectedFromScaleIn: value,
+			};
+			const command = new SetInstanceProtectionCommand(input);
+			const response = await autoScalingClient.send(command);
+			console.log('added scale-in protection', response);
```
philmcmahon, line 27: maybe this log message could include the instance id and remove line 18

packages/worker/src/index.ts
```diff
@@ -18,22 +18,28 @@ import {
 	getOrCreateContainer,
 } from './transcribe';
 import path from 'path';
+import { updateScaleInProtection } from './asg';
 
 const main = async () => {
 	const config = await getConfig();
+	const stage = config.app.stage;
+	const region = config.aws.region;
+	console.log('stage is: ', stage);
```
philmcmahon, line 27: maybe remove this?

philmcmahon:
This is looking great, a few minor comments

philmcmahon (approved):
scaling

Looks good!

PR #30, opened 2024-02-12
Add DynamoDB table and use it to store transcription data within output-handler lambda

srbd:
## What does this change?

We store the transcription text, original filename and email of user in dynamoDB table. The email sent from within this lambda will eventually contain a link to a google doc that will be populated using this data.

- Added dynamodb as a service in localstack in docker-compose
- Added dynamodb to `backend-common`
- Added transcriptTable dynamoDb table to the cdk and gave the outputHandler lambda read-write access to it
- output-handler lambda writes the transcription item to the table ahead of sending the email 

## How to test

Tested on CODE in the console using a test event


packages/output-handler/test/testMessage.ts
```diff
@@ -0,0 +1,22 @@
...
...\n  "TopicArn" : "arn:aws:sns:eu-west-1:745349931791:transcription-service-destination-topic-PROD",\n  "Message" : "{\\"id\\":\\"my-first-transcription\\",\\"transcriptionSrt\\":\\"1\\\\n00:00:00,000 --> 00:00:01,300\\\\n This is The Guardian.\\\\n\\\\n2\\\\n00:00:01,300 --> 00:00:06,300\\\\n [Music]\\\\n\\\\n\\",\\"languageCode\\":\\"en\\",\\"userEmail\\":\\"test@test.com\\",\\"originalFilename\\":\\"test.mp3\\"}",\n  "Timestamp" : "2024-02-08T11:10:11.014Z",\n  "SignatureVersion" : "1",\n  "Signature" : "N9Aj8kN72G2OAu5Y5xnZmdGaAjkHEzSCzRRQPMGLtIAPk9Pk1VwjrEc2TQZFSwt4L7Yv7n3XhY7CWexu/nYepbOn9ZG6scTtZ0tGQpsIPBYr7GPjjvvnRYjJtpHeAJzDRgfDAJWnFZhFTxU3XZfy+ZBuLl+86d2WRwe+U3+94AAUJUGr5m1eUGW8U2cj0hl4gPSv1czXwhu+PTyjOqAmWM2bFLWrYnuEIwrdOopBF+gAQGYVkGWbEgy+8GP4O9iMcBXMPyOoCOadNl599vbEX3z68A0kC4I51BSByYAkI8hqr71ONB+DwyYu87E8viu5G8IAq2BIT/jnmHk2fjdjfw==",\n  "SigningCertURL" : "https://sns.eu-west-1.amazonaws.com/SimpleNotificationService-60eadc530605d63b8e62a523676ef735.pem",\n  "UnsubscribeURL" : "https://sns.eu-west-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:eu-west-1:745349931791:transcription-service-destination-topic-PROD:ccd6477c-e3ec-4a0d-aa39-2c49eaa91a59"\n}',
```
philmcmahon, line 7: can you delete anything including the string 745349931791 from this message - I think for local testing we only need `Message`, everything else isn't needed - though you could just set all the values to `test` or something to give an idea of what it looks like on prod

.gitignore
```diff
@@ -7,3 +7,4 @@ node_modules/
 build
 **.js
 localstack/
+**/package-lock.json
```
philmcmahon, line 10: nice

philmcmahon (approved):
S06E04-K9YsO0t5-subtitled

packages/backend-common/src/dynamodb.ts
```diff
@@ -0,0 +1,53 @@
...
+	json: z.string(),
+});
+export const TranscriptionItem = z.object({
+	id: z.string(),
+	originalFilename: z.string(),
+	transcript: Transcript,
+	userEmail: z.string(),
+});
+
+export type TranscriptionItem = z.infer<typeof TranscriptionItem>;
```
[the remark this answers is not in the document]
srbd: ah! Did think about that but I think keeping related code close is good until it's shared

PR #31, opened 2024-02-13
Support export to google doc

philmcmahon:
## What does this change?
The main aim here is to allow users to export a transcript to a google doc. This involves the following features:
Client:
 - A new 'Export' page. This is the first React page that we have that is not just index.html. The way Next.js works is that it creates a separate page.html file for each page. As such I have had to update the production handler for client side pages -previously we always just returned index.html
 - A new 'ExportButton' component to handle export, waiting for an export, and linking to the generated google doc
 - `export.ts` service on the client which handles authentication with google to get permsissions to write to google drive
 - I've refactored `authFetch` to allow the entire request object to be overridden

Server:
 - A new /export endpoint to trigger the export and handle all the various error cases by returning an error with an appropriate status code
 - `googleDrive.ts` - this takes the authentication token shared by the client and uses it to actually create files in google drive
 - A new 'getTranscriptionItem' function to read values from dynamodb
 - Modification to how we serve the client side app - as discussed in the first point above

## How to test
You should be able to deploy to CODE/PROD and then use the test event on the output handler lambda (change the userEmail field) to get an email sent to you with a working export link.

Alternatively locally, you can run `npm run output-handler::start` to get a dummy item into your dynamodb table which you can then use to test the export on https://transcribe.local.dev-gutools.co.uk/export?transcriptId=my-first-transcription

export

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -100,14 +100,12 @@ const getApp = async () => {
 				);
 				const parsedItem = TranscriptionItem.safeParse(item);
 				if (parsedItem.success) {
-					console.log('parsed item', parsedItem.data);
 					const exportResult = await createTranscriptDocument(
 						config,
 						`${parsedItem.data.originalFilename} transcript`,
 						exportRequest.data.oAuthTokenResponse,
 						parsedItem.data.transcript.srt,
 					);
-					console.log('Export result', exportResult);
 					res.send({
 						documentId: exportResult,
 					});
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -147,10 +147,20 @@ const getApp = async () => {
 
 	app.use('/api', apiRouter);
 
+	const clientPages = ['export'];
+
 	if (runningOnAws) {
 		app.use(express.static('client'));
-		app.get('/*', (req, res) => {
-			res.sendFile(path.resolve(__dirname, 'client', 'index.html'));
+		app.get('/:page', (req, res) => {
+			if (req.params.page && !clientPages.includes(req.params.page)) {
+				res
+					.status(404)
+					.send(
+						`Endpoint not supported. Valid endpoints: /, /${clientPages.join(', /')}`,
+					);
+			}
+			const page = req.params.page ? `${req.params.page}.html` : 'index.html';
+			res.sendFile(path.resolve(__dirname, 'client', page));
 		});
 	} else {
 		if (emulateProductionLocally) {
@@ -159,7 +169,8 @@ const getApp = async () => {
 					path.resolve(__dirname, '..', '..', '..', 'packages/client/out'),
 				),
 			);
-			app.get('/*', (req: Request, res: Response) => {
+			app.get('/:page', (req: Request, res: Response) => {
+				const page = req.params.page ? `${req.params.page}.html` : 'index.html';
 				res.sendFile(
 					path.resolve(
 						__dirname,
@@ -167,7 +178,7 @@ const getApp = async () => {
 						'..',
 						'..',
 						'packages/client/out',
-						'index.html',
+						page,
 					),
 				);
 			});
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -87,44 +87,50 @@ const getApp = async () => {
 		checkAuth,
 		asyncHandler(async (req, res) => {
 			const exportRequest = TranscriptExportRequest.safeParse(req.body);
-			console.log(exportRequest);
 			const dynamoClient = getDynamoClient(
 				config.aws.region,
 				config.aws.localstackEndpoint,
 			);
-			if (exportRequest.success) {
-				const item = await getTranscriptionItem(
-					dynamoClient,
-					config.app.tableName,
-					exportRequest.data.id,
-				);
-				if (!item) {
-					const msg = `Failed to fetch item with id ${exportRequest.data.id} from database.`;
-					console.error(msg);
-					res.status(500).send(msg);
-					return;
-				}
-				const parsedItem = TranscriptionItem.safeParse(item);
-				if (parsedItem.success) {
-					const exportResult = await createTranscriptDocument(
-						config,
-						`${parsedItem.data.originalFilename} transcript`,
-						exportRequest.data.oAuthTokenResponse,
-						parsedItem.data.transcript.srt,
-					);
-					res.send({
-						documentId: exportResult,
-					});
-					return;
-				}
+			if (!exportRequest.success) {
+				const msg = `Failed to parse export request ${exportRequest.error.message}`;
+				console.error(msg);
+				res.status(400).send(msg);
+				return;
+			}
+			const item = await getTranscriptionItem(
+				dynamoClient,
+				config.app.tableName,
+				exportRequest.data.id,
+			);
+			if (!item) {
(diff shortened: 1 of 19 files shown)


packages/client/src/components/ExportButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+	return (
+		<>
+			<Script src="https://accounts.google.com/gsi/client" async></Script>
+			<button
+				onClick={async () => {
+					setLoading(true);
+					const response = await exportTranscript(token, transcriptId);
+					setLoading(false);
+					const parsedResponse = ExportResponse.safeParse(response);
+					if (parsedResponse.success) {
```
zekehuntergreen, line 41: `if (!parsedResponse.success)` shall we show an error?

srbd:
Tested on code, nice 🎉

packages/api/src/services/googleDrive.ts
```diff
@@ -0,0 +1,124 @@
...
+	const fileMetadata = {
+		name: folderName,
+		mimeType: 'application/vnd.google-apps.folder',
+	};
+	try {
+		// first see if there is already a folder matching folderName
+		const existingFolders = await drive.files.list({
+			q: `mimeType='application/vnd.google-apps.folder' and name ='${folderName}' and trashed=false`,
+			spaces: 'drive',
+		});
```
marjisound, line 19: Maybe get the mimetype from `fileMetadata.mimeType` rather than hard coding it?

packages/api/src/index.ts
```diff
@@ -63,6 +73,61 @@ const getApp = async () => {
...
+	apiRouter.post('/export', [
+		checkAuth,
+		asyncHandler(async (req, res) => {
+			const exportRequest = TranscriptExportRequest.safeParse(req.body);
+			console.log(exportRequest);
+			const dynamoClient = getDynamoClient(
+				config.aws.region,
+				config.aws.localstackEndpoint,
+			);
+			if (exportRequest.success) {
```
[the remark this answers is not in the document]
philmcmahon: good shout

packages/api/src/index.ts
```diff
@@ -63,6 +73,61 @@ const getApp = async () => {
...
+				if (parsedItem.success) {
+					const exportResult = await createTranscriptDocument(
+						config,
+						`${parsedItem.data.originalFilename} transcript`,
+						exportRequest.data.oAuthTokenResponse,
+						parsedItem.data.transcript.srt,
+					);
+					res.send({
+						documentId: exportResult,
+					});
```
marjisound, line 117: What if the `exportResult` is null? Do we not return error status in this case?

packages/client/src/components/UploadForm.tsx
```diff
@@ -41,6 +45,10 @@ export const UploadForm = ({ auth }: { auth: AuthState }) => {
 			`/signedUrl?${urlParams.toString()}`,
```
zekehuntergreen, line 45: think we need to change this to `/api/signedUrl` given the changes to `authFetch`

packages/api/src/index.ts
```diff
@@ -86,10 +151,20 @@ const getApp = async () => {
...
+		app.get('/:page', (req, res) => {
+			if (req.params.page && !clientPages.includes(req.params.page)) {
+				res
+					.status(404)
+					.send(
+						`Endpoint not supported. Valid endpoints: /, /${clientPages.join(', /')}`,
+					);
+			}
+			const page = req.params.page ? `${req.params.page}.html` : 'index.html';
+			res.sendFile(path.resolve(__dirname, 'client', page));
```
marjisound, line 167: I thought the nextjs had routing mechanism?  What was the limitation in this case?

packages/backend-common/src/dynamodb.ts
```diff
@@ -51,3 +55,23 @@ export const writeTranscriptionItem = async (
...
+		Key: {
+			id: itemId,
+		},
+	});
+	try {
+		const result = await client.send(command);
+		return result.Item;
+	} catch (error) {
+		console.error(`Failed to get item ${itemId} from dynamodb`, error);
+		return null;
```
marjisound, line 75: I think in general TypeScript guidelines recommends using `undefined` rather than null wherever possible. Do you think we can change this to be undefined?

packages/client/src/components/ExportButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { ExportResponse } from '@guardian/transcription-service-common';
+import { AuthContext } from '@/app/template';
+import Script from 'next/script';
+import { useSearchParams } from 'next/navigation';
+import { exportTranscript } from '@/services/export';
+
+const ExportButton = () => {
+	const auth = useContext(AuthContext);
+	const searchParams = useSearchParams();
+	const [docId, setDocId] = useState<string | undefined>(undefined);
```
marjisound, line 11: I think the initial `undefined` value is not needed, when there's no value, it'd be undefined;
`const [docId, setDocId] = useState<string | undefined>();`

packages/client/src/services/export.ts
```diff
@@ -0,0 +1,74 @@
...
+	const driveFileScope = 'https://www.googleapis.com/auth/drive.file';
+
+	const tokenResponse = await promiseInitTokenClient(
+		config.googleClientId,
+		driveFileScope,
+	);
+
+	const exportRequest: TranscriptExportRequest = {
+		id: transcriptId,
+		oAuthTokenResponse: tokenResponse,
```
marjisound, line 63: I'm getting type error in vs code on this. Is it looking fine on yours?

packages/client/src/helpers.ts
```diff
@@ -1,7 +1,20 @@
...
-		},
+export const authFetch = async (
+	url: string,
+	token?: string,
+	init?: RequestInit,
+): Promise<Response | null> => {
+	if (!token) {
+		console.error('Missing auth token');
+		return null;
+	}
```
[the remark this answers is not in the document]
philmcmahon: thanks, I've modified that. Required a small change to UploadForm

packages/client/src/services/export.ts
philmcmahon: Ah I had a ts-ignore comment which I mistakenly removed in a later commit, will add it back

packages/api/src/index.ts
philmcmahon: Thanks, I've tidied this up

packages/client/src/helpers.ts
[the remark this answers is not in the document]
zekehuntergreen: thanks!

packages/client/src/helpers.ts
```diff
@@ -1,7 +1,20 @@
...
+	const request = new Request(url, init);
+
+	request.headers.set('Authorization', `Bearer ${token}`);
+
+	const authRequest = new Request(request, {
+		headers: request.headers,
```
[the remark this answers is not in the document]
zekehuntergreen: even better

packages/api/src/index.ts
philmcmahon: It looks like nextjs deliberately generates different html files for each page for performance reasons rather than working as a single page app. https://nextjs.org/docs/app/building-your-application/deploying/static-exports

marjisound (approved):
Great work 🎉

PR #32, opened 2024-02-14
Add failure metric to worker and output-handler services

philmcmahon:
## What does this change?
This change adds the capability to post metrics to cloudwatch from our transcription service.

Initially, I have created one new metric called `Failure` to record failures during the transcription process. At the moment this metric will be updated in the following situations:
 - If the worker fails to parse an incoming SQS message
 - If the FFMPEG conversion or whisper transcription fails 
 - If the output-handler fails to either write to dynamo or send a notification email.

I wasn't sure whether to make the different failure scenarios different metrics or not (e.g. counting transcription failures in a different metric to 'notification failed' failures). I decided to put everything together with the idea that all we want to know is that something failed - we should then be able to go and look at the logs to work out exactly what failed.

The intention with this metric is that it should only ever be updated once per transcription job - so I'm only updating it in the top level handler functions.

At the moment there are no alarms attached to these metrics - I decided to hold off on that until PROD is stable so we don't get bored of alarms straight away.

## How to test
I tested this by deploying the branch and then kicking off a transcription, then logging onto the worker server running the transcription and calling `docker stop <container id>` to break the transcription process. I then verified that a the metric existed and the value had been updated - see `https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#metricsV2?graph=~(metrics~(~(~'TranscriptionService~'Failure))~view~'timeSeries~stacked~false~region~'eu-west-1~start~'-PT1H~end~'P0D~stat~'Maximum~period~300)&namespace=~'TranscriptionService`

In future, we should have good visibility of transcription failures. 

failure

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ build
 **.js
 localstack/
 **/package-lock.json
+**.DS_Store
--- a/package-lock.json
+++ b/package-lock.json
@@ -686,6 +686,513 @@
 				}
 			}
 		},
+		"node_modules/@aws-sdk/client-cloudwatch": {
+			"version": "3.512.0",
+			"resolved": "https://registry.npmjs.org/@aws-sdk/client-cloudwatch/-/client-cloudwatch-3.512.0.tgz",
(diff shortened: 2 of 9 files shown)


marjisound (approved):
Looks great 👌

PR #33, opened 2024-02-14
Use signed urls for transcription output

marjisound:
## What does this change?

This PR does the following:
- CDK - Creates a bucket for storing the transcripts output (srt, json & text)
- CDK - gives `s3:PutObject` permission on the bucket to the api lambda
- CDK - gives `s3:GetObject` permission on the bucket to the outputHandler lambda
- API - Creates S3 upload signed urls for the transcript output before sending the message to SQS
- API - Adds the signed urls to the SQS message
- Worker - Uploads the transcripts to the relevant S3 bucket using the signed urls
- Worker - Adds the S3 keys of the uploaded output files to the SNS message
- Output Handler - Downloads the transcripts from the output bucket using the keys provided by the SQS message
- Output Handler - Reads the transcripts files and adds their text to the object that goes to dynamo db

## How to test

Tested in code by doing the following:
- made an api call to `api/send-message` endpoint which is using hardcoded values for the message
- Waited for the instance to scale up
- Checked the logs in the instance and make sure it's all working fine 
- Waited for the email notification
- Checked dynamo db to make sure the record was added

## TODO:
For Prod release we need to add the new config parameter to parameter store

### DIFF
--- a/packages/cdk/lib/transcription-service.ts
+++ b/packages/cdk/lib/transcription-service.ts
@@ -262,13 +262,6 @@ export class TranscriptionService extends GuStack {
 				new GuPolicy(this, 'WorkerGetParameters', {
 					statements: [getParametersPolicy],
 				}),
-				new GuAllowPolicy(this, 'GetDeleteSourceMedia', {
-					actions: ['s3:GetObject', 's3:DeleteObject'],
-					resources: [
-						`${sourceMediaBucket.bucketArn}/*`,
-						`${outputBucket.bucketArn}/*`,
-					],
-				}),
 				new GuAllowPolicy(this, 'WriteToDestinationTopic', {
 					actions: ['sns:Publish'],
 					resources: [transcriptDestinationTopic.topicArn],
--- a/packages/cdk/lib/transcription-service.ts
+++ b/packages/cdk/lib/transcription-service.ts
@@ -262,6 +262,13 @@ export class TranscriptionService extends GuStack {
 				new GuPolicy(this, 'WorkerGetParameters', {
 					statements: [getParametersPolicy],
 				}),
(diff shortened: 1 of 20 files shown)


packages/common/src/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+interface UploadFailure {
+	isSuccess: false;
+	errorMsg?: string;
+}
+
+type UploadResult = UploadSuccess | UploadFailure;
+
+export const uploadToS3 = async (
```
zekehuntergreen, line 12: nice refactor 🛠️

packages/common/src/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+			};
+		}
+
+		return {
+			isSuccess: true,
+		};
+	} catch (error) {
+		console.error('upload error:', error);
+		return {
+			isSuccess: false,
```
philmcmahon, line 41: include the error message here?

packages/cdk/lib/transcription-service.ts
```diff
@@ -227,7 +264,10 @@ export class TranscriptionService extends GuStack {
 				}),
 				new GuAllowPolicy(this, 'GetDeleteSourceMedia', {
 					actions: ['s3:GetObject', 's3:DeleteObject'],
-					resources: [`${sourceMediaBucket.bucketArn}/*`],
+					resources: [
+						`${sourceMediaBucket.bucketArn}/*`,
+						`${outputBucket.bucketArn}/*`,
+					],
```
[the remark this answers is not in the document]
philmcmahon: I had a go at this before realising that the reading from the bucket is still done using the s3Key at the moment, let's save for your branch with new send message logic

PR #34, opened 2024-02-15
Block egress traffic from workers (except for AWS services)

philmcmahon:
## What does this change?
Prior to this PR we hadn't configured a security group for our transcription worker autoscaling group. As a result AWS applied the fallback default security group, which allows all outbound traffic.

For security reasons we'd like to block outbound traffic. Key considerations are:
 - Reassurance that transcript/audio data is not getting sent to some third party. Locking down egress traffic ensures this
 - If a users uploads some malware rather than some audio to transcribe, we want to limit the potential impact

This PR locks down egress traffic from the worker instances by applying a security group. However, in order to operate the worker does need access to some AWS services:
 - S3 - to download the worker deb package and transcript files
 - SQS - to fetch new transcript jobs
 - Autoscaling - to set e.g. scale-in protection
 - SNS - to post finished transcripts
 - Parameter store

This is achieved through the use of AWS Private Link/VPC endpoints which I've added here https://github.com/guardian/investigations-platform/pull/471. The S3 endpoint supports locking down to specific buckets/keys which we might want to consider at some point (otherwise in theory a malicious bit of code could upload to a random S3 bucket) but for now there's no restriction applied

This PR includes:
 - A new worker security group with egress rules for the new endpoints created in ttps://github.com/guardian/investigations-platform/pull/471. There are 2 rules - one that allows access to 'interface' vpc endpoints which is just a security group. For S3 we have to allow access to the 'managed prefix list' which appears to be global for our account - I've added as a parameter store parameter
 - I've updated the repository stack so that it allows the worker instances access to published containers **Note right now this isn't used as the containers are baked into the AMI** I thought this was a good idea because, following this change, the worker instances will no longer be able to access github packages to fetch the latest image.

## How to test
I've tested this on CODE - you can just deploy the branch and then check that transcription is still working.

gandalf being an egress rule

### DIFF
--- a/packages/cdk/lib/repository.ts
+++ b/packages/cdk/lib/repository.ts
@@ -30,6 +30,20 @@ export class TranscriptionServiceRepository extends GuStack {
 					'Deploy tools account id - needed to give AMIgo access to this repository',
 			},
 		);
+		const transcriptionWorkerRoleArnCODE = new GuStringParameter(
+			this,
+			'TranscriptionWorkerRoleArnCODE',
+			{
+				description: 'IAM role for the CODE transcription worker instances',
+			},
+		);
+		const transcriptionWorkerRoleArnPROD = new GuStringParameter(
+			this,
+			'TranscriptionWorkerRoleArnPROD',
+			{
+				description: 'IAM role for the PROD transcription worker instances',
+			},
+		);
 		const repository = new Repository(this, 'TranscriptionServiceRepository', {
 			repositoryName: `transcription-service`,
 			lifecycleRules: [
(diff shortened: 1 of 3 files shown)


packages/cdk/lib/repository.ts
```diff
@@ -43,7 +57,11 @@ export class TranscriptionServiceRepository extends GuStack {
 		});
 		repository.addToResourcePolicy(
 			new PolicyStatement({
-				principals: [new ArnPrincipal(githubActionsIAMRoleArn.valueAsString)],
+				principals: [
+					new ArnPrincipal(githubActionsIAMRoleArn.valueAsString),
+					new ArnPrincipal(transcriptionWorkerRoleArnCODE.valueAsString),
+					new ArnPrincipal(transcriptionWorkerRoleArnPROD.valueAsString),
```
philmcmahon, line 63: should be read only

marjisound (approved):
Amazing work and thanks for explaining it in a call

PR #35, opened 2024-02-15
Add message to task queue when file is uploaded

zekehuntergreen:
## What does this change?
- updates the `transcribe-file` api endpoint (renamed from `send-message`) so that it 
  - takes an  `s3Key` parameter 
  - creates a presigned url to download the file with the given key in the uploads bucket
  - adds a message to the task queue (with non-test data) including this url so that the worker can use it to get the file
- change to client so that when the user submits the form, after the file is uploaded to s3, `transcribe-file` is called.
- change to worker so that it uses this presigned url rather than having access to the entire bucket. This change is important because it will allow third parties to integrate with the transcription service without our having to change any of the worker permissions. 

## How to test
- can be tested locally by starting api, client, worker
- testing in CODE

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -63,7 +63,7 @@ const getApp = async () => {
 		}),
 	]);
 
-	apiRouter.post('/send-message', [
+	apiRouter.post('/transcribe-file', [
 		checkAuth,
 		asyncHandler(async (req, res) => {
 			const userEmail = req.user?.email;
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -9,15 +9,14 @@ import passport from 'passport';
 import { Request, Response } from 'express';
 import {
 	getConfig,
-	getSignedUrl,
+	getSignedUploadUrl,
 	getSQSClient,
 	sendMessage,
 	isFailure,
-	getDownloadSignedUrl,
+	getSignedDownloadUrl,
 } from '@guardian/transcription-service-backend-common';
 import {
 	ClientConfig,
-	SignedUrlQueryParams,
 	TranscriptExportRequest,
 	sendMessageRequestBody,
 } from '@guardian/transcription-service-common';
@@ -77,7 +76,7 @@ const getApp = async () => {
 			}
 
 			const s3Key = body.data.s3Key;
-			const signedUrl = await getDownloadSignedUrl(
+			const signedUrl = await getSignedDownloadUrl(
 				config.aws.region,
 				config.app.sourceMediaBucket,
 				s3Key,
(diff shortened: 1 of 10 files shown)


packages/api/src/index.ts
```diff
@@ -61,21 +63,33 @@ const getApp = async () => {
...
+			const userEmail = req.user?.email;
+			console.log(req.body);
+			const body = sendMessageRequestBody.safeParse(req.body);
+			if (!body.success || !userEmail) {
+				console.error(body, userEmail);
+				res.status(422).send('missing request params');
+				return;
+			}
+
+			const s3Key = body.data.s3Key;
```
zekehuntergreen, line 78: need to check that current user was the one to upload the file with this key

packages/backend-common/src/s3.ts
```diff
@@ -57,28 +70,52 @@ export const getFile = async (
...
+	await downloadS3Data(body, destinationPath, key);
+	return destinationPath;
+};
+
+const downloadS3Data = async (
+	data: Readable,
+	destinationPath: string,
+	key: string,
+) => {
+	const body = ReadableBody.parse(data);
```
marjisound, line 100: It seems that we have already parsed the sream into `ReadableBody` on line 90. Probably this is redundant here?

marjisound (commented):
Looks great, I just had 2 minor questions

packages/backend-common/src/s3.ts
```diff
@@ -32,15 +31,29 @@ export const getSignedUrl = (
...
+			Key: id,
 			Metadata: {
 				'user-email': userEmail,
-				'file-name': fileName,
 			},
 		}),
 		{ expiresIn }, // override default expiration time of 15 minutes
 	);
 
+export const getSignedDownloadUrl = async (
```
philmcmahon, line 42: slightly out of the scope of this pr but I think worth doing - could generate the s3 client at the top level in the various index.ts files and then pass it through to this function and getSignedUploadurl.

You could potentially reduce code duplication even more by having just one function that either takes a GetObjectCommand or a PutObjectCommand

packages/api/src/index.ts
```diff
@@ -147,25 +161,21 @@ const getApp = async () => {
...
+			const s3Key = uuid4();
+			const presignedS3Url = await getSignedUploadUrl(
 				config.aws.region,
 				config.app.sourceMediaBucket,
 				req.user?.email ?? 'not found',
-				queryParams.data.fileName,
 				60,
 				true,
+				s3Key,
 			);
```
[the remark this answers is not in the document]
zekehuntergreen: from conversation in standup last week - we've decided we're ok with orphaned files in the media bucket when the upload to the bucket succeeds but the POST to transcribe-file fails. The user will just have to try uploading again

marjisound (approved):
Looks great 🎉

PR #39, opened 2024-02-20
Add worker-capacity-manager lambda to scale ASG to match queue length

srbd:
## What does this change?

Adds a lambda to run every minute and scale the ASG according to queue length ={invisible + visible messages total}.

This also removes the old scaling logic which lived under `packages/cdk/target_tracking_scaling_policy`

## How to test

Tested on CODE. Log messages and activity in the ASG can see scaling capacity up and down according to messages placed on queue


packages/worker-capacity-manager/package.json
```diff
@@ -0,0 +1,24 @@
...
+		"start": "STAGE=DEV nodemon src/index.ts"
+	},
+	"keywords": [],
+	"author": "",
+	"license": "ISC",
+	"dependencies": {
+		"@guardian/transcription-service-common": "1.0.0",
+		"@guardian/transcription-service-backend-common": "1.0.0",
+		"@aws-sdk/client-sqs": "^3.514.0",
+		"@aws-sdk/client-auto-scaling": "^3.514.0"
```
philmcmahon, line 17: client-sqs and client-autoscaling are included in backend-common so don't need to be explicity added here

packages/worker-capacity-manager/src/sqs.ts
```diff
@@ -0,0 +1,28 @@
...
+				Number(attributes.ApproximateNumberOfMessages) +
+				Number(attributes.ApproximateNumberOfMessagesNotVisible)
+			);
+		}
+	} catch (error) {
+		console.log("Couldn't get queue length", error);
+		throw error;
+	}
+	throw new Error('unable to retrieve sqs message count attributes');
+};
```
marjisound, line 28: When can this throw occur? Is it only when `attributes` is undefined? If that's the case, could it not be as an else at the point where we check the `attributes`?

marjisound (approved):
Looks great, just left a small suggestion 🎉

packages/worker-capacity-manager/src/asg.ts
```diff
@@ -0,0 +1,21 @@
...
+	capacity: number,
+) => {
+	try {
+		const command = new SetDesiredCapacityCommand({
+			AutoScalingGroupName: asgGroupName,
+			DesiredCapacity: capacity,
+		});
+		await asgClient.send(command);
+	} catch (error) {
+		console.log("Couldn't set desired capacity", error);
```
zekehuntergreen, line 18:
```suggestion
		console.error("Couldn't set desired capacity", error);
```

packages/worker-capacity-manager/src/sqs.ts
```diff
@@ -0,0 +1,28 @@
...
+		const response = await sqsClient.send(command);
+		const attributes = response.Attributes;
+		if (attributes) {
+			return (
+				Number(attributes.ApproximateNumberOfMessages) +
+				Number(attributes.ApproximateNumberOfMessagesNotVisible)
+			);
+		}
+	} catch (error) {
+		console.log("Couldn't get queue length", error);
```
zekehuntergreen, line 24:
```suggestion
		console.error("Couldn't get queue length", error);
```

packages/cdk/lib/transcription-service.ts
```diff
@@ -514,5 +522,46 @@ export class TranscriptionService extends GuStack {
 			exportName: `WorkerRoleArn-${props.stage}`,
 			value: workerRole.roleArn,
 		});
+
+		const workerCapacityManagerLambda = new GuLambdaFunction(
```
zekehuntergreen, line 526: non-blocking: I see guardian CDK has a [`GuScheduledLambda`](https://guardian.github.io/cdk/classes/patterns.GuScheduledLambda.html) which looks like shorthand for the GuLambdaFunction + schedule Rule and might be worth using here if we can? (Not sure it saves us that many lines of code though)

packages/worker/src/asg.ts
```diff
@@ -1,22 +1,17 @@
-import {
-	AutoScalingClient,
-	SetInstanceProtectionCommand,
-} from '@aws-sdk/client-auto-scaling';
+import { SetInstanceProtectionCommand } from '@aws-sdk/client-auto-scaling';
 import { readFile } from '@guardian/transcription-service-backend-common';
+import { getASGClient } from '@guardian/transcription-service-backend-common/src/asg';
```
zekehuntergreen, line 3: encapsulation of backend-common is a bit better if we import from `transcription-service-backend-common`. Then we can reorganize modules in transcription-service-backend-common without changing this import.
```suggestion
import { readFile, getASGClient } from '@guardian/transcription-service-backend-common';
```

packages/worker-capacity-manager/package.json
```diff
@@ -0,0 +1,24 @@
+{
+	"name": "worker-capacity-manager",
+	"version": "1.0.0",
+	"description": "",
```
zekehuntergreen, line 4: It would be good to include a readme.md on what this package does and the other approaches we tried.
(could be another PR if you want to get this one out)

packages/worker-capacity-manager/src/sqs.ts
srbd: Intellij gave me bad advice! I did have it this way originally, changed it back!

packages/cdk/lib/transcription-service.ts
srbd: ah yes I've even used these before - will do in another PR perhaps in order not to block

packages/worker-capacity-manager/package.json
srbd: Yep good suggestion

PR #40, opened 2024-02-20
Add some styling to UI

philmcmahon:
## What does this change?
This PR make the transcribe tool look a bit nicer:

https://github.com/guardian/transcription-service/assets/3606555/0dae7130-1698-4272-8764-c0c8e06897e7

I've used the component library made by the tailwind people - https://flowbite.com/docs/. There are a lot of extra dependencies - could definitely do something lighter weight than this for the sake of a few buttons!!! But for now this works

Things that would be nicer:
 - get rid of [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) and flash of 'login required' message somehow - maybe with loading spinner rather than showing login message

### Why not elastic UI?
Elastic UI only supports typescript 4 unfortunately and I couldn't face the impact downgrading might have on our Next.js build

## How to test
I've tested this on CODE, it seems to look ok.

### DIFF
--- a/packages/client/src/app/layout.tsx
+++ b/packages/client/src/app/layout.tsx
@@ -1,4 +1,5 @@
 import './globals.css';
+import React from 'react';
 export default function RootLayout({
 	// Layouts must accept a children prop.
 	// This will be populated with nested layouts or pages
@@ -8,7 +9,20 @@ export default function RootLayout({
 }) {
 	return (
 		<html lang="en">
-			<body>{children}</body>
+			<body>
+				<header className="bg-white shadow">
+					<div className="mx-auto max-w-7xl px-4 py-6 sm:px-6 lg:px-8">
+						<h1 className="text-3xl font-bold tracking-tight text-gray-900">
+							Guardian Transcribe Tool
+						</h1>
+					</div>
+				</header>
+				<main>
+					<div className="mx-auto max-w-7xl py-6 sm:px-6 lg:px-8">
+						{children}
+					</div>
+				</main>
+			</body>
 		</html>
 	);
 }
--- a/packages/client/src/app/layout.tsx
+++ b/packages/client/src/app/layout.tsx
@@ -13,7 +13,7 @@ export default function RootLayout({
 				<header className="bg-white shadow">
 					<div className="mx-auto max-w-7xl px-4 py-6 sm:px-6 lg:px-8">
 						<h1 className="text-3xl font-bold tracking-tight text-gray-900">
-							Guardian Transcribe Tool
+							Guardian Transcription Tool
 						</h1>
 					</div>
 				</header>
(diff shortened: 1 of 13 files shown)


packages/client/src/components/UploadForm.tsx
```diff
@@ -1,96 +1,158 @@
...
-			console.error('Failed to call transcribe-file');
-			return false;
-		}
-		return true;
-	};
+	if (!token) {
+		return (
+			<InfoMessage message={'Login required'} status={RequestStatus.Failed} />
+		);
+	}
```
zekehuntergreen, line 60: can token ever be undefined here, given we're checking it in `template.tsx`? do we just need to convince typescript of this?

packages/client/src/components/UploadForm.tsx
```diff
@@ -1,96 +1,158 @@
...
 
-		const success = await uploadFileAndTranscribe(maybeFileInput);
-		setStatus(success);
-		if (success) {
-			maybeFileInput.value = '';
+		for (const file of maybeFileInput.files) {
+			const result = await uploadFileAndTranscribe(file, token);
+			if (!result) {
+				setStatus(RequestStatus.Failed);
+				return;
```
marjisound, line 129: About the return statement, is it because if there are multiple files, the first file that failes to upload, we want to stop the upload?

marjisound (approved):
What a lovely UI we have now 🤩 great work

packages/client/src/app/layout.tsx
```diff
@@ -8,7 +9,20 @@ export default function RootLayout({
 }) {
 	return (
 		<html lang="en">
-			<body>{children}</body>
+			<body>
+				<header className="bg-white shadow">
+					<div className="mx-auto max-w-7xl px-4 py-6 sm:px-6 lg:px-8">
+						<h1 className="text-3xl font-bold tracking-tight text-gray-900">
+							Guardian Transcribe Tool
```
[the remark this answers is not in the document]
philmcmahon: I think transcription is better 👍🏻

packages/client/src/app/template.tsx
```diff
@@ -23,16 +23,25 @@ export default function Template({ children }: { children: React.ReactNode }) {
 
 	useEffect(() => {
 		const browserHistory = createBrowserHistory();
-		initAuth(setAuth, browserHistory);
+
+		const newAuth = initAuth(browserHistory);
+		console.log('initAuth', newAuth);
```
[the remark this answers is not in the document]
philmcmahon: oops no!

packages/client/src/components/UploadForm.tsx
philmcmahon: That is what was happening but I think it was a mistake -we should try the next file rather than just exiting

packages/client/src/components/UploadForm.tsx
philmcmahon: At the moment it can be undefined because in the AuthState type in template.tsx it is initialised as undefined. I think it would be good to change that though as otherwise every page we add will need to do this additional check. Something for another bit of work/PR though I think

PR #43, opened 2024-02-22
Worker polls for transcription tasks every 30 seconds

zekehuntergreen:
## What does this change?
- changes transcriptions workers so that they poll the task queue every 30 seconds, rather than only polling once, when they're started up.
- this should help to stop workers from getting into a state where they are doing no work but preventing other workers from taking their place

## How to test
- [x] test on code

### DIFF
--- a/packages/worker/src/index.ts
+++ b/packages/worker/src/index.ts
@@ -29,10 +29,9 @@ import {
 } from '@guardian/transcription-service-backend-common/src/metrics';
 import { SQSClient } from '@aws-sdk/client-sqs';
 import { SNSClient } from '@aws-sdk/client-sns';
+import { setTimeout } from 'timers/promises';
 
 const POLLING_INTERVAL_SECONDS = 30;
-// avoid two tasks being processed by a single worker simultaneously
-let transcriptionTaskSemaphoreTaken = false;
 
 const main = async () => {
 	const config = await getConfig();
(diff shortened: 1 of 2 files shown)


packages/worker/src/index.ts
```diff
@@ -26,41 +27,91 @@ import {
...
 	const numberOfThreads = config.app.stage === 'PROD' ? 16 : 2;
 
-	const client = getSQSClient(region, config.aws.localstackEndpoint);
-	const message = await getNextMessage(client, config.app.taskQueueUrl);
+	if (transcriptionTaskSemaphoreTaken) {
+		console.log(`transcription task in progress. skipping`);
+		return;
+	}
+	console.log(
+		`worker polling for transcription task. Poll count = ${pollCount}`,
```
zekehuntergreen, line 95: what's a better way to say poll count? 🤔

marjisound: retry count?

zekehuntergreen: hmm I feel like retry implies a previous _failed_ attempt. Whereas polling and receiving 0 messages isn't really a failure or an error case.

packages/worker/src/index.ts
```diff
@@ -26,41 +27,91 @@ import {
...
+		pollTranscriptionQueue(pollCount, sqsClient, snsClient, metrics, config);
+	}, POLLING_INTERVAL_SECONDS * 1000);
+};
+
+const releaseSemaphoreAndRemoveScaleInProtection = async (
+	region: string,
+	stage: string,
+) => {
+	transcriptionTaskSemaphoreTaken = false;
+	await updateScaleInProtection(region, stage, false);
```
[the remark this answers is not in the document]
zekehuntergreen: good call! I've changed the error message wording

packages/worker/src/index.ts
marjisound: Yes that's right

zekehuntergreen:
> ```
> // try and get message, if there is one do transcription
> // regardless of what happens, setTimeout(30 seconds, start again)
> ```

That's much simpler than how I was doing it! No need to bother with the semaphore and after a transcription, the next time we poll the queue will be 30 seconds later, rather than anywhere from 0-30 seconds later.

PR #46, opened 2024-02-27
Move task to dead letter queue when conversion to wav fails

zekehuntergreen:
## What does this change?
- updates CDK to allow worker instances to write sqs messages to the dead letter task queue
- when workers fail when processing a file that can't be converted to a wav, move the sqs message to the dead letter queue

## How to test
- [x] tested in CODE by attempting to transcribe a jpeg
- [x] transcription still working as expected for an mp3

### DIFF
--- a/packages/backend-common/src/sqs.ts
+++ b/packages/backend-common/src/sqs.ts
@@ -114,7 +114,7 @@ const sendMessage = async (
 			errorMsg: 'Missing message ID',
 		};
 	} catch (e) {
-		const msg = `Failed to send job ${messageBody}`;
+		const msg = `Failed to send message ${messageBody}`;
 		console.error(msg, e);
 		return {
 			status: SQSStatus.Failure,
@@ -225,10 +225,8 @@ export const moveMessageToDeadLetterQueue = async (
 	// succeeds but the delete from the task queue fails
 	const sendResult = await sendMessage(client, deadLetterQueueUrl, messageBody);
 	if (sendResult.status == SQSStatus.Failure) {
-		const errorMessage = 'Failed to send message to dead letter queue.';
-		console.error(errorMessage, sendResult.error, sendResult.errorMsg);
 		// rethrow exception, let another worker retry
-		throw Error(errorMessage);
+		throw Error('Failed to send message to dead letter queue');
 	}
 	// if the delete command throws an exception, it will be caught by
 	// deleteMessage and logged. Another worker will reprocess the message in
--- a/packages/backend-common/src/sqs.ts
+++ b/packages/backend-common/src/sqs.ts
@@ -28,6 +28,10 @@ interface ReceiveSuccess {
 	message?: Message;
 }
 
+interface DeleteSuccess {
+	status: SQSStatus.Success;
+}
+
 interface SQSFailure {
 	status: SQSStatus.Failure;
 	error?: unknown;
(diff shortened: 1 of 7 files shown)


packages/backend-common/src/sqs.ts
```diff
@@ -179,18 +191,47 @@ export const deleteMessage = async (
...
+	// succeeds but the delete from the task queue fails
+	const sendResult = await sendMessage(client, deadLetterQueueUrl, messageBody);
+	if (sendResult.status == SQSStatus.Failure) {
+		// rethrow exception, let another worker retry
+		throw Error('Failed to send message to dead letter queue');
 	}
+	// if the delete command throws an exception, it will be caught by
+	// deleteMessage and logged. Another worker will reprocess the message in
+	// the main task queue and we'll have a duplicate in the dead letter queue.
+	await deleteMessage(client, taskQueueUrl, receiptHandle);
```
philmcmahon, line 234: great documentation here, good shout to see if there was an atomic operation available

packages/worker/src/index.ts
```diff
@@ -136,6 +143,28 @@ const pollTranscriptionQueue = async (
 		);
 
 		const ffmpegResult = await convertToWav(containerId, fileToTranscribe);
+		if (ffmpegResult === undefined) {
+			// when ffmpeg fails to transcribe, move message to the dead letter
+			// queue
+			if (config.app.stage != 'DEV' && config.app.deadLetterQueueUrl) {
+				console.log(
+					`moving message with message id ${taskMessage.MessageId} to dead letter queue`,
```
philmcmahon, line 151: maybe 'ffmpeg failed, moving message....'

packages/worker/src/transcribe.ts
```diff
@@ -127,7 +127,7 @@ export const convertToWav = async (
 		};
 	} catch (error) {
 		console.log('ffmpeg failed error:', error);
-		throw error;
+		return;
```
philmcmahon, line 130: what's the motivation for returning void here vs undefined. Is it because there are other failure states that return void? I'm not bothered, but I think elsewhere we have tended towards returning undefined in failure situtations. It is a shame typescript doesn't have a built in Either type!

philmcmahon (approved):
One minor comment and a query but this is looking good to me

dead letters

packages/backend-common/src/sqs.ts
```diff
@@ -179,18 +191,47 @@ export const deleteMessage = async (
...
 	} catch (error) {
-		console.error(`Failed to delete message ${receiptHandle}`, error);
-		throw error;
+		const errorMsg = `Failed to delete message ${receiptHandle}`;
+		console.error(errorMsg, error);
+		return {
+			status: SQSStatus.Failure,
+			error,
+			errorMsg,
+		};
```
philmcmahon, line 212: 👍🏻 good to have a proper failure return type

packages/worker/src/transcribe.ts
zekehuntergreen: I'd always used void and undefined as return types interchangeably but [the top answer to this question](https://stackoverflow.com/questions/58885485/why-does-typescript-have-both-void-and-undefined) is a good explanation of the distinction. I'll change to undefined since that's what we're doing elsewhere.

philmcmahon:
changes look good 👍🏻

PR #47, opened 2024-02-27
Improve logging using winston

marjisound:
## What does this change?

This PR adds `winston` logging library (only for serverside logging) in order to send structured data for logging and also improve the logging behaviour. 
The following improvements were done: 
- adds userEmail, fileName, message id to the API logging where the message is successfully pushed to SQS
- adds userEmail, fileName, file duration and message id to the WORKER logging where message is successfully sent to SNS
- adds userEmail, fileName, message id to the output handler logging where email is successfully sent
- adds message id and user email to all worker loggings (after sqs message is parsed up to when the processing is finished for that message)
- adds @timestamp to all messages - if this field doesn't exist in the log message, the DevEx tool that ships logs to ELK automatically generates timestamp but it seems that the  sametimestamp is generated for a batch of messages and place the messages out of order in kibana
- adds `level` to all logs - this make troubleshooting easier by filtering for error logs
- We had this issue where the logs were being seperated by new lines (`\n`) and it was hard to make sense of them in kibana. Using winston json format, fixed this issue. 

## How to test

Tested in CODE and locally

### DIFF
--- a/packages/client/src/components/UploadForm.tsx
+++ b/packages/client/src/components/UploadForm.tsx
@@ -25,7 +25,7 @@ const uploadFileAndTranscribe = async (file: File, token: string) => {
 	}
 
 	const uploadStatus = await uploadToS3(body.data.presignedS3Url, blob);
-	if (!uploadStatus) {
+	if (!uploadStatus.isSuccess) {
 		console.error('Failed to upload to s3');
 		return false;
 	}
--- a/packages/worker/src/transcribe.ts
+++ b/packages/worker/src/transcribe.ts
@@ -50,8 +50,8 @@ const runSpawnCommand = (
 				stderr: stderr.join(''),
 				code: code || undefined,
 			};
-			logger.info(result.stdout);
-			logger.info(result.stderr);
+			logger.info(result.stdout.replace('\n', ' '));
+			logger.info(result.stderr.replace('\n', ' '));
 			if (code === 0) {
 				resolve(result);
 			} else {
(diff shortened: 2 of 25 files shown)


srbd (commented):
Do you have an example of the log format before and after for easy comparison?

marjisound:
> Do you have an example of the log format before and after for easy comparison?

The best way to see it in action is to check in kibana. But I can share one of the logs in my terminal:
```
{"@timestamp":"2024-02-28T13:58:06.888Z","fileDuration":"6","filename":"random-file-marji.wav","id":"618ba0b7-4405-4517-8725-c8fd6242b18e","level":"info","message":"Worker successfully transcribed the file and sent notification to sns","timestamp":"2024-02-28T13:58:06.888Z","userEmail":"marjan.kalanaki@guardian.co.uk"}
```
These will be the fields that are sent from the app, but there are more fields in kibana based on where the logs are coming from (lambda or EC2). This is an example of the fields 
image

packages/client/src/components/UploadForm.tsx
```diff
@@ -25,7 +25,7 @@ const uploadFileAndTranscribe = async (file: File, token: string) => {
 	}
 
 	const uploadStatus = await uploadToS3(body.data.presignedS3Url, blob);
-	if (!uploadStatus) {
+	if (!uploadStatus.isSuccess) {
```
zekehuntergreen, line 28: 👍

packages/worker/src/transcribe.ts
```diff
@@ -32,12 +34,10 @@ const runSpawnCommand = (
 		const stdout: string[] = [];
 		const stderr: string[] = [];
 		cp.stdout.on('data', (data) => {
-			console.log(data.toString());
```
zekehuntergreen, line 35: is this where we were logging transcript text?

philmcmahon (approved):
Such a good improvement I'm looking forwards to having more readable logs! I like the setCommonMetadata function - should make it much easier to trace what's happened with a particular transcription.

logs before they're logs

PR #48, opened 2024-02-27
Store transcripts in s3 rather than dynamo

philmcmahon:
## What does this change?
Dynamo string files have a [max size of 400kb](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ServiceQuotas.html). To work round this let's store transcript data in S3 instead.

This was an easy change, as transcripts are already in S3! We need to warn users though as whilst dynamo data persists forever at the moment, we are expiring items in our output s3 bucket after 7 days.

 The changes in this PR are:
 - fix a bug where we were uploading the json output as text
 - warn users about transcript expiry
 - return helpful error on export suggesting expiry as the reason for a failed export
 - adjust dynamo transcript type to reflect that it doesn't actually contain the trainscript text
 - remove the code to download transcripts from the output handler
 - add a new s3 getObjectText to download an object straight into memory 
 - use getObjectText in the export handler to fetch the file from S3

## How to test
Deploy this change and run a transcript - if it works then the new logic going via s3 is working. I've tested this on CODE.


packages/api/src/index.ts
```diff
@@ -162,18 +166,28 @@ const getApp = async () => {
...
 			}
-			const parsedItem = TranscriptionItem.safeParse(item);
+			const parsedItem = TranscriptionDynamoItem.safeParse(item);
 			if (!parsedItem.success) {
 				const msg = `Failed to parse item ${exportRequest.data.id} from dynamodb. Error: ${parsedItem.error.message}`;
 				console.error(msg);
 				res.status(500).send(msg);
 				return;
 			}
+			const transcriptText = await getObjectText(
```
zekehuntergreen, line 176: I think it's worth considering forgoing dynamo altogether and just using s3 to save ourselves a request here and in the output handler. the s3 object has the user's email in its metadata and can be found with just a message id.

zekehuntergreen: I guess a counterargument is that storing transcriptKeys in dynamo gives us a bit more flexibility to change the file structure in s3

philmcmahon: This is a good point - the main reason we added dynamo was to store the transcript text. However, I think it's useful to have all the various bits of metadata associated with a transcript in one place as e.g. we can use it for audting or for building a nicer UI. We could dump the entire json blob, metadata and all in S3 as well though.

I think it's worth having a DB, if we wanted to implement a progress bar for example, it would be much nicer doing it with dynamo as it allows you to update a single property rather than having to overwrite the whole file. 

I've worked on a project before where they didn't bother with a database and just used S3 for everything. It definitely worked - the tools has been in use for years - but debugging was always a faff as you were constantly downloading random files rather than being able to query a database. Dynamo is particularly nice as we can just see all the data in the browser https://eu-west-1.console.aws.amazon.com/dynamodbv2/home?region=eu-west-1#item-explorer?maximize=true&operation=SCAN&table=transcription-service-PROD

packages/backend-common/src/dynamodb.ts
```diff
@@ -31,19 +31,19 @@ export const Transcripts = z.object({
 
 export type Transcripts = z.infer<typeof Transcripts>;
 
-export const TranscriptionItem = z.object({
+export const TranscriptionDynamoItem = z.object({
 	id: z.string(),
 	originalFilename: z.string(),
-	transcripts: Transcripts,
+	transcriptKeys: Transcripts,
```
marjisound, line 37: Would it not be better if we change the type name to something like `TranscriptKeys`?

packages/backend-common/src/s3.ts
```diff
@@ -58,7 +58,8 @@ export const getSignedDownloadUrl = async (
 		{ expiresIn }, // override default expiration time of 15 minutes
 	);
 
-export const getFile = async (
+// for larger files (such as media files)- stream the file to disk
+export const streamObjectToFile = async (
```
marjisound, line 62: This is not used anywhere right now, are you thinking it'll be used in the future?

packages/api/src/index.ts
```diff
@@ -162,18 +166,28 @@ const getApp = async () => {
...
+				s3Client,
+				config.app.transcriptionOutputBucket,
+				parsedItem.data.transcriptKeys.text,
+			);
+			if (!transcriptText) {
+				const msg = `Failed to export transcript - it is possible your transcript has expired. Please re-upload the file and try again.`;
+				res.status(500).send(msg);
```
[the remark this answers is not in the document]
philmcmahon: Good shout, I'm just trying to write this so we return a 410 if the file isn't there, but a 500 for any other S3 error, but otherwise will just return a 410 for everything

marjisound (commented):
Looks great 👍

packages/backend-common/src/s3.ts
philmcmahon: it is used in packages/backend-common/src/s3.ts

marjisound: Yea it is used in `getFileFromS3` function but getFileFromS3 doesn't seem to be used anywhere

packages/backend-common/src/s3.ts
```diff
@@ -101,10 +123,26 @@ export const getObjectText = async (
...
+		return {
+			status: S3Status.Success,
+			text: Buffer.concat(chunks).toString('utf-8'),
+		};
+	} catch (error: unknown) {
+		if (error instanceof Error) {
+			if (error.name === 'NoSuchKey') {
+				return {
+					status: S3Status.Failure,
+					statusCode: 410,
```
zekehuntergreen, line 106: non-blocking: `statusCode` isn't used in `api/src/index.ts` yet. Maybe it's better to keep it that way and remove `statusCode` from `GetObjectTextFailure`? That way it's clearer, reading `index.ts` which cases result in which status codes?

packages/backend-common/src/s3.ts
philmcmahon: sorry you're absolutely right, I had assumed this must be used when we donwload the media files but that's been replaced by `getObjectWithPresignedUrl`

packages/api/src/index.ts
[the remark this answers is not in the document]
zekehuntergreen: thanks!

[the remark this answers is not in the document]
philmcmahon: I've updated this (still needs testing on CODE) - we now:
 - 404 if user doesn't own the file
 - 410 if the file doesn't exist on S3
 - 500 for any other errror downloading from S3

packages/backend-common/src/s3.ts
philmcmahon: Removed in https://github.com/guardian/transcription-service/pull/48/commits/2389593f125beca55762b079f60c05e02e346499

packages/output-handler/src/index.ts
```diff
@@ -142,6 +94,7 @@ const processMessage = async (event: unknown) => {
 };
 
 const handler: Handler = async (event) => {
+	console.log(event);
```
zekehuntergreen, line 97: logger.info?

zekehuntergreen (approved):
looks good!

copying_big_file

packages/output-handler/src/index.ts
philmcmahon: I missed a few console. issues - resolved in https://github.com/guardian/transcription-service/pull/48/commits/db192aa2bc2486fec2a807641056aa17f5f33f2c (removed this particularly log line as I just needed it to get an updated message for `testMessage.ts` which is used to run the output-handler locally

PR #49, opened 2024-02-28
Add alarms for failures and cost warnings

philmcmahon:
## What does this change?
This adds 3 cloudwatch alarms to the PROD transcription stack:
 - dead letter alarm - if a message goes into the dead letter queue it suggests a user upload has failed to process, so we alarm on that condition
 - transcription failure alarm - we already have a metric added in https://github.com/guardian/transcription-service/pull/32 to track failures. This adds an alarm to that metric
 - Long running instances alarm - this will alarm if we have more than 0 instances running for a period of 5 hours (this might need some tweaking to get it right)

Depends on https://github.com/guardian/investigations-platform/pull/480

## How to test
I've tested these alarms by:
 - manually publishing a message to the dead letter queue
 - killing docker part way through a transcription
 - running some transcriptions for a long time to check the long runnning instances alarm works

### DIFF
--- a/packages/cdk/lib/transcription-service.ts
+++ b/packages/cdk/lib/transcription-service.ts
@@ -654,7 +654,7 @@ export class TranscriptionService extends GuStack {
 					threshold: 1,
 					comparisonOperator:
 						ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
-					evaluationPeriods: 5 * 12, // 5 hours as metric has period of 5 minutes
+					evaluationPeriods: 2, // testing 5 * 12, // 5 hours as metric has period of 5 minutes
 					actionsEnabled: true,
 					alarmDescription:
 						'There has been more than 1 worker instance running for 5 hours - this will have significant cost implications. Please check that all running workers are doing something useful.',
--- a/packages/cdk/lib/transcription-service.ts
+++ b/packages/cdk/lib/transcription-service.ts
@@ -37,6 +37,13 @@ import {
 	GroupMetrics,
 	SpotAllocationStrategy,
 } from 'aws-cdk-lib/aws-autoscaling';
+import {
+	Alarm,
+	ComparisonOperator,
+	Metric,
+	TreatMissingData,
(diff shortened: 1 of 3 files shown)


packages/cdk/lib/transcription-service.ts
```diff
@@ -576,5 +583,86 @@ export class TranscriptionService extends GuStack {
...
+					threshold: 1,
+					comparisonOperator:
+						ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
+					evaluationPeriods: 5 * 12, // 5 hours as metric has period of 5 minutes
+					actionsEnabled: true,
+					alarmDescription:
+						'There has been more than 1 worker instance running for 5 hours - this will have significant cost implications. Please check that all running workers are doing something useful.',
+					treatMissingData: TreatMissingData.IGNORE,
+				}),
+			];
```
marjisound, line 659: I'm trying to get my head around this last alarm. From what I understand from this alarm, during 5 hours, if there are different instances that are being scaled out (not  only 1 instance running for 5 hours), this alarm would still be triggered right?

philmcmahon: Yes - every 5 minutes we check the minimum number of instances that were running during that 5 minute period, if it is >0 for every 5 minute period over 5 hours then the alarm triggers.

I did this because it's a lot easier than tracking whether one specific instance has been running for more than 5 hours. I still think this is a pretty unlikely situation - because the alarm will reset as soon as we have 0 workers for 60 seconds.

philmcmahon:
Following a discussion in standup I've stopped the worker from incrementing the failure metric for general failures as these will be picked up by the dead letter alarm https://github.com/guardian/transcription-service/pull/49/commits/d8a3b18b56f97674a4351cbf6cb6e55795d88dcc

packages/cdk/lib/transcription-service.ts
marjisound: Thanks @philmcmahon for explaining, it makes sense 👍

marjisound (approved):
Great work

giphy (1)

PR #52, opened 2024-02-28
Use the transcription id to provide a uniqe MessageGroupId

philmcmahon:
## What does this change?
Another go at https://github.com/guardian/transcription-service/pull/50 - we need to set a unique MessageGroupId for each transcription job going to SQS - otherwise we can't process transcription jobs in parallel.

I decided to use the transcript ID for the MessageGroupId, seems as good as anything

### DIFF
--- a/packages/backend-common/src/sqs.ts
+++ b/packages/backend-common/src/sqs.ts
@@ -100,7 +100,7 @@ const sendMessage = async (
 			new SendMessageCommand({
 				QueueUrl: queueUrl,
 				MessageBody: messageBody,
-				MessageGroupId: 'api-transcribe-request',
+				MessageGroupId: id,
 			}),
 		);
 		console.log(`Message sent. Message id: ${result.MessageId}`);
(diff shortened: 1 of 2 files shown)


packages/backend-common/src/sqs.ts
```diff
@@ -86,13 +86,14 @@ export const generateOutputSignedUrlAndSendMessage = async (
...
 	};
-	return await sendMessage(client, queueUrl, JSON.stringify(job));
+	return await sendMessage(client, queueUrl, JSON.stringify(job), id);
 };
 
 const sendMessage = async (
 	client: SQSClient,
 	queueUrl: string,
 	messageBody: string,
+	id: string,
```
zekehuntergreen, line 96: on main, is this being used at the moment? 🤔

marjisound: We just checked with Zeke, and you may need a rebase from main. Not sure why github is not showing the diff with the latest main 🤔

philmcmahon: Thanks for this review, I have fixed it now. Apologies for these rushed PRs - should have properly tested to start with but got overexcited.

zekehuntergreen (approved):
id

PR #53, opened 2024-02-28
Use spot instances

philmcmahon:
## What does this change?
In order to reduce the costs for our transcription service, let's use spot instances for the transcription workers.

A lot of the infrastructure for using spot instances was added in https://github.com/guardian/transcription-service/pull/4 - that PR has quite a bit of detail. What this change does is actually 'switch on' spot instances by setting `onDemandPercentageAboveBaseCapacity` to 0. With that change, all instances will be launched as spot instances. This means

 - instances may take a little longer to start up
 - instances may be terminated with 2 minutes notice - see [spot interruption notice](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-instance-termination-notices.html)

Regarding the startup time, we should monitor this (will be easier once we have more structured logs etc) but I don't anticipate this being a problem - in our tests so far instances took a few seconds longer than ondemand to start.

The termination is a bigger issue - which is addressed in this PR. We've added a job which polls the instance metadata service (using IMDSV1 - we should switch to IMDSV2 at some point) to check for a termination notice. If it finds one then we know that the instance will be terminated in the next 2 minutes.

The most likely scenario for this to happen is when we are mid transcription, likely with much more than 2 minutes left to go. For this scenario, we simply reduce the visibility timeout on the job currently being transcribed to 2 minutes, so that after our instance is terminated the job can be quickly picked up by another instance.

On the off chance that we are able to complete our transcription in less than 2 minutes then it would be nice to not throw away all the work. However, we don't want to 'half finish' a transcription - e.g. send a message to the ouput handler but not delete the transcription job from the task queue. To prevent this from happening we've got a nice `if` condition that checkes a new `INTERRUPT_TIME` variable to decide whether it's safe to start uploading a finished transcript (we allow 20s which should be more than enough to upload a few text files to S3 and notify SNS/SQS.

## How to test
You can test this by triggering a spot instance interruption as outlined here https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/initiate-a-spot-instance-interruption.html 

Test cases tried on CODE:
 - Run a transcription, check it works
 - Run a transcription, immediately trigger an interrupt, verify that the work gets picked up by a new instance and the transcription still completes
 - Run a transcription, interrupt when transcription is almost complete, check that transcription finishes succesfully

Money saving! 

make it rain

### DIFF
--- a/packages/worker/src/spot-termination.ts
+++ b/packages/worker/src/spot-termination.ts
@@ -12,21 +12,38 @@ export const checkSpotInterrupt = async (
 	queueUrl: string,
 ) => {
 	const url = 'http://169.254.169.254/latest/meta-data/spot/instance-action';
-	const result = await fetch(url);
-	if (result.status === 200) {
-		const json = await result.json();
-		if (json.action === 'terminate') {
-			setInterruptionTime(new Date(json.time));
-			logger.warn('Spot instance termination detected');
-			// Interrupt warning occurs 2 minutes before termination
-			const receiptHandle = getCurrentReceiptHandle();
-			if (!receiptHandle) {
+	try {
+		const result = await fetch(url);
+		if (result.status === 200) {
+			const json = await result.json();
+			if (json.action === 'terminate') {
+				const interruptionTime = new Date(json.time);
+				setInterruptionTime(interruptionTime);
+				logger.warn('Spot instance termination detected');
+				// Interrupt warning occurs 2 minutes before termination
+				const receiptHandle = getCurrentReceiptHandle();
+				if (!receiptHandle) {
+					return;
+				}
+				const secondsUntilTermination =
+					(interruptionTime.getTime() - new Date().getTime()) / 1000;
+				try {
+					await changeMessageVisibility(
+						client,
+						queueUrl,
+						receiptHandle,
+						secondsUntilTermination,
+					);
+				} catch (e) {
+					// we don't care if the visibility change fails - it will just delay the message being
+					// picked up by a new worker, so do nothing here
+				}
+				// once the interrupt warning has happened, we don't need to keep checking
 				return;
 			}
-			await changeMessageVisibility(client, queueUrl, receiptHandle, 110);
-			// once the interrupt warning has happened, we don't need to keep checking
-			return;
 		}
+	} catch (e) {
+		console.error('Error during spot termination check', e);
 	}
 	setTimeout(
 		() => checkSpotInterrupt(client, queueUrl),
--- a/packages/worker/src/spot-termination.ts
+++ b/packages/worker/src/spot-termination.ts
@@ -19,6 +19,8 @@ export const checkSpotInterrupt = async (
 			console.warn('Spot instance termination detected');
 			// Interrupt warning occurs 2 minutes before termination
 			await changeMessageVisibility(client, queueUrl, receiptHandle, 110);
+			// once the interrupt warning has happened, we don't need to keep checking
+			return;
 		}
 	}
 	setTimeout(
--- a/packages/worker/src/spot-termination.ts
+++ b/packages/worker/src/spot-termination.ts
@@ -1,13 +1,15 @@
-import { changeMessageVisibility } from '@guardian/transcription-service-backend-common';
+import {
+	changeMessageVisibility,
+	logger,
+} from '@guardian/transcription-service-backend-common';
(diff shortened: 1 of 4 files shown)


philmcmahon:
Some issues I found when testing:
 - We don't want to keep checking for interrupt after we have received an interrupt - fixed in https://github.com/guardian/transcription-service/pull/53/commits/c505e5c1af7eee8de2830a4f8abcf70aabf94f80
 - We only want to start the spot interrupt checker once - not once per loop - fixed in https://github.com/guardian/transcription-service/pull/53/commits/afe01db52a6d5e7855aa2d4decd12a388c023d42

packages/worker/src/spot-termination.ts
```diff
@@ -0,0 +1,35 @@
...
+import { getCurrentReceiptHandle, setInterruptionTime } from './index';
+
+const CHECK_FREQUENCY = 10;
+
+export const checkSpotInterrupt = async (
+	client: SQSClient,
+	queueUrl: string,
+) => {
+	const url = 'http://169.254.169.254/latest/meta-data/spot/instance-action';
+	const result = await fetch(url);
```
zekehuntergreen, line 15: do we need to catch network errors? seems like we need to be careful with error handling here since if there's an exception we'll stop polling for metadata

packages/worker/src/spot-termination.ts
```diff
@@ -0,0 +1,35 @@
...
+		const json = await result.json();
+		if (json.action === 'terminate') {
+			setInterruptionTime(new Date(json.time));
+			logger.warn('Spot instance termination detected');
+			// Interrupt warning occurs 2 minutes before termination
+			const receiptHandle = getCurrentReceiptHandle();
+			if (!receiptHandle) {
+				return;
+			}
+			await changeMessageVisibility(client, queueUrl, receiptHandle, 110);
```
zekehuntergreen, line 26: Might be worth considering relying on termination time from instance metadata rather than the two minutes from docs. this way message visibility and `INTERRUPTION_TIME` agree
```suggestion
			await changeMessageVisibility(client, queueUrl, receiptHandle, new Date(json.time) - new Date());
```

packages/worker/src/spot-termination.ts
```diff
@@ -0,0 +1,35 @@
...
+			// once the interrupt warning has happened, we don't need to keep checking
+			return;
```
zekehuntergreen, line 28: non-blocking: `changeMessageVisibility` rethrows exceptions which shouldn't matter because there's nothing left to do. It might be worth either changing its signature or wrapping it in a try / catch anyway. This way if we change this function to do more stuff after `changeMessageVisibility` it'll be guaranteed to happen

packages/worker/src/spot-termination.ts
```diff
@@ -0,0 +1,35 @@
...
+		}
```
zekehuntergreen, line 29: it might be too complicated to change in this PR but at the moment I think when a very unlucky job is picked up by three successive workers which are interrupted before finished transcription, we will put the message on the dead letter queue without notifying the user. 
We could use the two minutes before termination to send a failure message. 
Maybe for the moment, just having the alarm when the message is added to the dead letter queue is enough and we can find out as people use the service how common this case actually is.

zekehuntergreen:
This is looking great! Just a few comments

packages/worker/src/spot-termination.ts
philmcmahon: good question, `169.254.169.254` is a local IP address for the instance metadata service so should be fairly reliable but I will wrap in a try/finally to be sure

packages/worker/src/spot-termination.ts
```diff
@@ -0,0 +1,52 @@
...
+) => {
+	const url = 'http://169.254.169.254/latest/meta-data/spot/instance-action';
+	try {
+		const result = await fetch(url);
+		if (result.status === 200) {
+			const json = await result.json();
+			if (json.action === 'terminate') {
+				const interruptionTime = new Date(json.time);
+				setInterruptionTime(interruptionTime);
+				logger.warn('Spot instance termination detected');
```
zekehuntergreen, line 22: shall we log termination time?
```suggestion
				logger.warn(`Spot instance interruption detected. Interruption time: ${interruptionTime}`);
```

zekehuntergreen (approved):
thumsup

PR #55, opened 2024-02-29
Fix serverless express deprecation notice

zekehuntergreen:
## What does this change?
The api logs are full of this deprecation notice from serverless express
```
message | [DEPRECATION NOTICE] You're using the deprecated `serverlessExpress({...}).handler({...})` method. This will be removed in a future version of @codegenie/serverless-express. Instead, simply return `serverlessExpress({...})` as your handler.
```
This PR makes the recommended change so that logs are less noisy


philmcmahon (approved):
Nice 

handle it

PR #56, opened 2024-02-29
Extract timing and language data from transcript output

philmcmahon:
## What does this change?
This pr uses some regular expressions to extract useful information out of the whisper.cpp output. Currently, we extract language code, total time and load time (time taken to load the model).

## How to test
I've tested this on CODE - here's an example log line https://logs.gutools.co.uk/s/investigations/app/discover#/doc/3501e690-a763-11eb-882c-5b432fbcaa69/logstash-investigations-2024.02.29?id=Acc69Y0BLE-XvQ1FY3Mn


marjisound (approved):
Amazing work, it's great you also retrieved total and load time 🎉

PR #57, opened 2024-02-29
Tidy up UI - don't flash 'login required' message

philmcmahon:
## What does this change?
I noticed we flash a 'login required' message on most pages even when the user is logged in. This change adds a loading spinner when we're reading the value from local storage, and pulls the call to createBrowserHistory to the top of the component - which appears to speed things up a bit though would be good to check with @marjisound if you think this will work.

Before:

https://github.com/guardian/transcription-service/assets/3606555/5e930c3f-3319-47b8-b354-6a86d6731879

After:

https://github.com/guardian/transcription-service/assets/3606555/d939b2a7-92d7-40fa-b844-8e57c2a91b32

### DIFF
--- a/packages/client/src/app/template.tsx
+++ b/packages/client/src/app/template.tsx
@@ -12,7 +12,6 @@ export const AuthContext: React.Context<AuthState> = createContext({});
 
 export default function Template({ children }: { children: React.ReactNode }) {
 	const [authLoading, setAuthLoading] = useState(true);
-	const browserHistory = createBrowserHistory();
 	const [auth, setAuth] = useState<AuthState>(initialState);
 
 	useEffect(() => {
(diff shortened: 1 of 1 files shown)


marjisound (approved):
Thanks Phil for doing this. It looks perfect and certainly a great improvement 🎉

philmcmahon:
Thanks @marjisound  - just in case the PR description was confusing, I orginally moved the browser history call but then realised it broke things so ended up doing something simpler with fewer code change - which was what was eventually merged

marjisound:
> Thanks @marjisound - just in case the PR description was confusing, I orginally moved the browser history call but then realised it broke things so ended up doing something simpler with fewer code change - which was what was eventually merged

Ah I see, thanks for explaining 😄

PR #58, opened 2024-02-29
Set app tag of api to transcription-service-api

philmcmahon:
## What does this change?
To make it easier to identify in the logs, let's give transcription-service-api a more precise tag.

## How to test
I created a change set just to check this wasn't going to explode anything, looks ok:
Screenshot 2024-02-29 at 17 10 40


marjisound (approved):
This is great 👍

PR #59, opened 2024-03-01
Make file duration a number in logs

philmcmahon:
## What does this change?
Tiny change - this would allow us to have a table visualisation in our ELK cluster of e.g. average file length, max file length etc


marjisound (approved):
Nice 👍

PR #62, opened 2024-03-05
Bump transcription file signed download url timeout to 7 days

philmcmahon:
## What does this change?
If a transcription takes more than 2 hours and gets killed halfway through by e.g. spot instance rotation then it will need to be re-downloaded.

With that in mind, let's set a longer expiry for the s3 download url


zekehuntergreen (approved):
👍

PR #63, opened 2024-03-05
Allow users to select transcription language

zekehuntergreen:
## What does this change?
- Adds a dropdown to the form to let the user choose the language of the audio in their media file(s)
  - the language options are manually copy / pasted from [whisper.cpp's code](https://github.com/ggerganov/whisper.cpp/blob/25d313b38b1f562200f915cd5952555613cd0110/whisper.cpp#L251) 
  - the dropdown defaults to 'auto detect language', which is the current behaviour
  - some wording is added to the form to tell the user that choosing a language improves transcription accuracy
- The language code is passed in the request to api/transcribe-file, in the SQS message, to the worker which passes it as an argument to whisper
Screenshot 2024-03-06 at 09 35 49

- two other small changes
  - don't call `checkSpotInterrupt` in DEV
  - remove unused `retryCount` from task queue message. We rely on ApproximateReceiveCount

### Types
The tricky part of this change was getting the types and zod parsers as narrow as possible without defining the map of language code to language in more than one place.
I couldn't find a way to do this without two type assertions in common/types.ts and common/languages.ts but they are relatively contained.

## How to test
- [x] tested in CODE

- transcription accuracy should be better when language selected rather than auto-detected


philmcmahon (approved):
LGTM 👍🏻 

babel fish

hoyla:
Gorgeous

PR #65, opened 2024-03-08
Remove notifications to digital.investigations

philmcmahon:
## What does this change?
We no longer need to receive an email every time a transcription is completed as we can use the [usage dashboard](https://metrics.gutools.co.uk/d/BjGFOolnk/investigations-tools-usage?orgId=1&from=1709683200000&to=1709942399000) to analyse usage of the tool.


marjisound (approved):
LGTM

zekehuntergreen (approved):

PR #68, opened 2024-03-15
Remove SNS

philmcmahon:
## What does this change?
Initially when we started this project we had an idea to have the worker instances post to a single SNS topic which then feeds into multiple sqs topics for different services (giant, composer etc). Later, we decided that we didn't want to mix up output destined for giant with output for the transcription-service-output-handler - we shouldn't share data with services that aren't supposed to be consuming it, 

Having SNS provided some benefit of easy email notifications when a worker had finished a transcription, though again for privacy/spam reasons we've switched this off https://github.com/guardian/transcription-service/pull/65

However, when trying to integrate the transcription service with giant I found that using SNS added some complexity:
 - having to create a giant output topic 
 - having to add the SNS sdk to giant

So, given that we're not getting much benefit out of SNS, this PR removes it from the transcription service

## How to test
I've tested this on CODE and verified that transcription still works


packages/backend-common/src/configHelpers.ts
```diff
@@ -28,7 +28,6 @@ export const getParameters = async (
 		} while (nextToken);
 
 		if (parameters) {
-			logger.info('Fetched parameters from Parameter Store');
```
zekehuntergreen, line 31: 🙏

packages/worker/src/index.ts
```diff
@@ -308,9 +296,9 @@ const pollTranscriptionQueue = async (
 			taskMessage.Attributes?.ApproximateReceiveCount || defaultReceiveCount,
 		);
 		if (receiveCount >= MAX_RECEIVE_COUNT) {
-			publishTranscriptionOutputFailure(
-				snsClient,
-				config.app.destinationTopicArns.transcriptionService,
+			await publishTranscriptionOutputFailure(
```
zekehuntergreen, line 299: good call - it's probably important to await the result of this before updating scale in protection

packages/backend-common/src/config.ts
```diff
@@ -79,7 +79,7 @@ export const getConfig = async (): Promise<TranscriptionConfig> => {
 	const destinationTopic = findParameter(
```
zekehuntergreen, line 79: `destinationQueueUrls` ?

packages/output-handler/src/sqs-event-types.ts
```diff
@@ -3,15 +3,10 @@ import { stringToJSONSchema } from './zod-string-to-json';
 import { TranscriptionOutput } from '@guardian/transcription-service-common';
 
 const SQSMessageBody = z.object({
-	MessageId: z.string(),
-	Timestamp: z.string(),
-	Message: stringToJSONSchema.pipe(TranscriptionOutput),
+	messageId: z.string(),
+	body: stringToJSONSchema.pipe(TranscriptionOutput),
```
zekehuntergreen, line 7: non-blocking: I'm not sure I understand these changes. output-handler is still reading off of the same sqs queue, right? Why has the shape of the message changed?

zekehuntergreen (approved):
This looks good!
It would be great to update the architecture diagram with these changes when they're merged

clean_up

packages/output-handler/src/sqs-event-types.ts
philmcmahon: This was unexpected for me as well - I had to work out the structure of the new message by printing it out but I think think previously you'd have an sqs message wrapped around an sns notification, whereas now you've got an sqs message wrapped around the message itself.

PR #71, opened 2024-03-20
Enforce selection of media file language

zekehuntergreen:
## What does this change?
- changes transcription form so that there is no default language chosen, and the user must either select a language or 'Auto-detect language'.
- fixes the form on mobile by changing the language picker from a flowbite-react `Dropdown` element to a react `Select` element. The former didn't work on some mobile devices. When it was selected, the dropdown opened upward (regardless of `placement` property) and couldn't be navigated by swiping or scrolling.
- improves handling of form state by moving from a query selector to get files on form submission, to keeping track of that state with a `useState` hook

## Before
auto-detect language is the default, language picker doesn't work

https://github.com/guardian/transcription-service/assets/17057932/ec16286c-5e80-40b0-9744-260ffc4aaf8e

## after 
no default language, language picker does work

https://github.com/guardian/transcription-service/assets/17057932/b05d5c25-62fb-405c-a10f-50673aaf765a

## How to test
- [x] tested in CODE

- fewer mistakenly detected transcriptions
- users can use language picker on mobile

### DIFF
--- a/packages/client/src/components/UploadForm.tsx
+++ b/packages/client/src/components/UploadForm.tsx
@@ -75,7 +75,6 @@ const updateFileStatus = (
 
 export const UploadForm = () => {
 	const [status, setStatus] = useState<RequestStatus>(RequestStatus.Ready);
-	const [errorMessage, setErrorMessage] = useState<string | undefined>();
 	const [files, setFiles] = useState<FileList | null>(null);
 	const [uploads, setUploads] = useState<Record<string, RequestStatus>>({});
 	const [mediaFileLanguageCode, setMediaFileLanguageCode] = useState<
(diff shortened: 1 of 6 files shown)


packages/client/src/components/UploadForm.tsx
```diff
@@ -157,22 +163,24 @@ export const UploadForm = () => {
...
+		if (mediaFileLanguageCode === undefined) {
+			setLanguageCodeValid(false);
 			return;
 		}
+
+		// the required property on the file input should prevent the form from
+		// being submitted without any files selected. Need to confirm this in
+		// order to narrow the type of files
+		if (files === null || files.length === 0) {
+			console.error('form submitted without any files');
```
philmcmahon, line 178: since we don't have anything shipping client side logs anywhere, I'm not sure it's worth logging here

philmcmahon (approved):
This looks like a well thought out improvement, makes sense!

itsibitzi (approved):
🚢

PR #72, opened 2024-03-21
Update export page wording to encourage users to allow pop-ups

zekehuntergreen:
## What does this change?
Updates messages while the google auth window should be open and when authentication fails. We've had two users email us about auth errors that were caused by the browser blocking the pop-up.

https://github.com/guardian/transcription-service/assets/17057932/73fa09ed-2fbe-40e6-b26a-406c9e0359e7


philmcmahon (approved):
pop up

PR #73, opened 2024-03-28
Add option to choose the transcript format

marjisound:
## What does this change?

This PR adds options on the transcripts format when user is exporting the transcripts to a google document. The options are text and srt. It also adds validation so user can not submit the export unless they choose the format. 

The api is also updated to use the `transcriptFormat` value from the query parameters in order to provide the relevant text format.

ezgif-6-e8096e3ad7

## How to test

Tested locally and in CODE

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -189,10 +189,14 @@ const getApp = async () => {
 				res.status(404).send(`Transcript not found`);
 				return;
 			}
+
+			const transcriptS3Key =
+				parsedItem.data.transcriptKeys[exportRequest.data.transcriptFormat];
+			console.log(`transcriptS3Key: ${transcriptS3Key}`);
 			const transcriptText = await getObjectText(
 				s3Client,
 				config.app.transcriptionOutputBucket,
-				parsedItem.data.transcriptKeys.text,
+				transcriptS3Key,
 			);
 			if (isS3Failure(transcriptText)) {
 				if (transcriptText.failureReason === 'NoSuchKey') {
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -192,7 +192,6 @@ const getApp = async () => {
 
 			const transcriptS3Key =
 				parsedItem.data.transcriptKeys[exportRequest.data.transcriptFormat];
-			console.log(`transcriptS3Key: ${transcriptS3Key}`);
 			const transcriptText = await getObjectText(
 				s3Client,
 				config.app.transcriptionOutputBucket,
--- a/packages/client/src/components/ExportButton.tsx
+++ b/packages/client/src/components/ExportButton.tsx
@@ -1,5 +1,8 @@
 import React, { useContext, useState } from 'react';
-import { ExportResponse } from '@guardian/transcription-service-common';
+import {
+	ExportResponse,
+	TranscriptFormat,
+} from '@guardian/transcription-service-common';
 import { AuthContext } from '@/app/template';
 import Script from 'next/script';
 import { useSearchParams } from 'next/navigation';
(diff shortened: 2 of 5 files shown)


packages/client/src/components/ExportButton.tsx
```diff
@@ -69,9 +85,18 @@ const ExportButton = () => {
 	}
 
 	const exportHandler = async () => {
+		if (!transcriptFormat) {
+			console.log(`transcript format value is ${transcriptFormat}`);
```
philmcmahon, line 89: remove

packages/client/src/components/ExportButton.tsx
```diff
@@ -26,6 +35,13 @@ const ExportButton = () => {
 			<InfoMessage message={'not logged in'} status={RequestStatus.Failed} />
 		);
 	}
+
+	const transcriptFormatDescription: Record<TranscriptFormat, string> = {
+		srt: 'SRT format (Time Coded)',
+		text: 'TEXT format',
+		json: 'JSON format',
```
philmcmahon, line 42: I would remove the word 'format' from these since there is a label on the dropdown. I think TEXT should be title case 'Text' as I imagine many journalists won't be aware of TXT/ 'text' format - 'Text' seems clearer

packages/api/src/index.ts
```diff
@@ -189,10 +189,14 @@ const getApp = async () => {
 				res.status(404).send(`Transcript not found`);
 				return;
 			}
+
+			const transcriptS3Key =
+				parsedItem.data.transcriptKeys[exportRequest.data.transcriptFormat];
+			console.log(`transcriptS3Key: ${transcriptS3Key}`);
```
philmcmahon, line 195: remove

philmcmahon (approved):
Looks great! Just one comment about not repeating the word 'format' but you might have already discussed that so feel free to ignore

packages/client/src/components/ExportButton.tsx
```diff
@@ -98,9 +123,55 @@ const ExportButton = () => {
...
+						value={TranscriptFormat.TEXT}
+						onClick={() => {
+							setTranscriptFormat(TranscriptFormat.TEXT);
+							setTranscriptFormatValid(true);
+						}}
+					>
+						{transcriptFormatDescription[TranscriptFormat.TEXT]}
+					</Dropdown.Item>
+					<Dropdown.Divider />
+					<Dropdown.Item
```
zekehuntergreen, line 156: Should we make srt the default option in the dropdown? From journalist feedback it seems like this is the format they want most of the time.

packages/client/src/components/ExportButton.tsx
```diff
@@ -26,6 +35,13 @@ const ExportButton = () => {
 			<InfoMessage message={'not logged in'} status={RequestStatus.Failed} />
 		);
 	}
+
+	const transcriptFormatDescription: Record<TranscriptFormat, string> = {
+		srt: 'SRT format (Time Coded)',
```
zekehuntergreen, line 40: another formatting suggestion: 
could we make this `SRT (time-coded)` or `SRT (with time codes)`

packages/client/src/components/ExportButton.tsx
marjisound: This is more of a UX point. I think by enforcing to choose one of the formats we are also letting user know that there are multiple formats to choose from, but if there's a format chosen by default, user might not even notice there are other formats. We can chat with Luke and see what he says too.

philmcmahon: I think Text might still be the most popular format - I suspect if we set SRT as the default initially then everyone would have emailed asking for a timecode-free version

zekehuntergreen: good points! let's go with no default

PR #78, opened 2024-04-04
Worker logs file language, output handler records completion date

zekehuntergreen:
## What does this change?
- our tools usage dashboard is broken since switching from autodetected transcription language to user-specified languages in https://github.com/guardian/transcription-service/pull/71 since it relied on autodetected language. This change logs the user-specified language as well as the autodetected one so that the dashboard can be fixed.
- the output handler records completion date of transcription in dynamodb so that we have more specific reporting on usage numbers


marjisound (approved):
LGTM

PR #80, opened 2024-04-18
Use whisper.cpp 1.5.5

philmcmahon:
## What does this change?
See https://github.com/ggerganov/whisper.cpp/releases/tag/v1.5.5 - latest release of whisper.cpp. We might want to wait a few days before merging this in case there's a rapid follow up bugfix release.


marjisound (approved):
Looks good, thanks for doing it

PR #81, opened 2024-04-23
Update CODEOWNERS - investigations not investigations-classic

philmcmahon:
## What does this change?
Now that I&R is just one big team we don't need the subteam any more 

we are one


srbd (approved):
I don't remember if I can attach a gif this way

PR #100, opened 2024-08-30
Support translation

philmcmahon:
## What does this change?
The library we are using for transcription - [whisper.cpp](https://github.com/ggerganov/whisper.cpp) - supports translation to english.

This PR modifies the transcription tool to make use of this feature. A new checkbox is added to the submission form:

Screenshot 2024-08-30 at 14 00 12

This value is sent through to the API. If true, then the API will send 2 messages rather than 1 to the queue. The extra message will have the same id but with `-translation` appended. 

The worker then proceeds as normal, but when it finishes and uploads the output text, if it's a translation then `-translation` is appended to each of the json/srt/txt filenames.

The output handler modifies the email subject and message if it's a translation to indicate this:
Screenshot 2024-08-30 at 14 03 11

The end result is that the user receives 2 emails. Whilst it would be nicer to have just one email with links to both docs, this approach means that they get the transcript as soon as possible, and we don't have to deal with any kind of state management of waiting for one or the other outputs.

## How to test
Tested on CODE

### DIFF
--- a/packages/output-handler/src/index.ts
+++ b/packages/output-handler/src/index.ts
@@ -66,6 +66,7 @@ const handleTranscriptionSuccess = async (
 		},
 		userEmail: transcriptionOutput.userEmail,
 		completedAt: new Date().toISOString(),
+		isTranslation: transcriptionOutput.isTranslation,
 	};
 
 	try {
--- a/packages/output-handler/src/index.ts
+++ b/packages/output-handler/src/index.ts
@@ -27,18 +27,23 @@ const successMessageBody = (
 	transcriptId: string,
 	originalFilename: string,
 	rootUrl: string,
+	isTranslation: boolean,
 ): string => {
 	const exportUrl = `${rootUrl}/export?transcriptId=${transcriptId}`;
 	return `
-		<h1>Transcript for ${originalFilename} ready</h1>
+		<h1>Transcript ${isTranslation ? 'english translation ' : ''}for ${originalFilename} ready</h1>
 		<p>Click <a href="${exportUrl}">here</a> to export to a google doc.</p>
 		<p><b>Note:</b> transcripts will expire after 7 days. Export your transcript to a doc now if you want to keep it. </p>
 	`;
 };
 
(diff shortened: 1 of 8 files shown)


hoyla:
Exciting!

packages/output-handler/src/index.ts
```diff
@@ -74,11 +80,12 @@ const handleTranscriptionSuccess = async (
 			sesClient,
 			config.app.emailNotificationFromAddress,
 			transcriptionOutput.userEmail,
-			`Transcription complete for ${transcriptionOutput.originalFilename}`,
+			`Transcription ${transcriptionOutput.isTranslation ? 'english translation ' : ''}complete for ${transcriptionOutput.originalFilename}`,
```
zekehuntergreen, line 83: what do you think about this?
```suggestion
			`${transcriptionOutput.isTranslation ? 'English translation' : 'Transcription'} complete for ${transcriptionOutput.originalFilename}`,
```

zekehuntergreen: If it's a translation email we don't want subject to be "Transcription english translation complete..." right?

zekehuntergreen: and same for failure subject and message body?

philmcmahon: hmm I was thinking that it's technically a translation of the transcription, but "transcription english translation is a bit of a mouthfull! I think your suggestion is clearer - it's kinda implicit that it's a transcription

philmcmahon: updated in https://github.com/guardian/transcription-service/pull/100/commits/47678134842921b085e54defa5dc173899df2c1b

hoyla:
When you get the email saying the transcription (and translation?) is complete, and "Click here to export to a google doc", would it open both docs? Am wondering how people might get to each of the two documents

philmcmahon:
Yep there are 2 emails which will result in 2 google docs if they export them both
Screenshot 2024-08-30 at 16 27 12

marjisound (approved):
This is great use of 10%, nice work

PR #103, opened 2024-10-02
enabling transcription integration for giant

marjisound:
## What does this change?

This PR is enabling transcription service to process messages coming from third parties service (in particular 'Giant'). 
- The config model for `destinationQueueUrls` is updated so that we can now dynamically get the destination value based on if the message is coming from transcription or giant. 
- the worker message output is now published to the relevant service based on the value of `transcriptDestinationService` in the input message
- Added a local dead letter queue for the local task input queue to make it consistent with our code/prod environments
- Added a giant output queue and giant output dead letter queue. In code/prod these queues belong to giant, but in local setup, because adding 2 localstack would have add complication, we decided to have them created as part of the transcription service localstack. 
- Thanks to @philmcmahon for adding logic for translation if the message is coming from giant 🎉

The relevant PRs for this change and the order they need to be released are as followed:
1- https://github.com/guardian/investigations-platform/pull/521
2- current PR
3- https://github.com/guardian/giant/pull/239

The following SSM parameters were created for code but should also be created for prod:
- /CODE/investigations/GiantTranscriptionOutputQueueArn
- /CODE/investigations/transcription-service/destinationQueueUrls/giant
## How to test

Tested this both locally and in code

### DIFF
--- a/packages/worker/src/transcribe.ts
+++ b/packages/worker/src/transcribe.ts
@@ -229,6 +229,8 @@ const transcribeAndTranslate = async (
 			false,
 		);
 
+		// we only run language detection once,
+		// so need to override the detected language of future whisper runs
 		transcription.metadata.detectedLanguageCode = metadata.detectedLanguageCode;
 		const translation =
 			languageCode === 'en'
--- a/packages/worker/src/transcribe.ts
+++ b/packages/worker/src/transcribe.ts
@@ -222,12 +222,14 @@ const transcribeAndTranslate = async (
 		const dlParams = whisperParams(true, whisperBaseParams.wavPath);
 		const { metadata } = await runWhisper(whisperBaseParams, dlParams);
 		const languageCode =
-			languageCodes.find((c) => c === metadata.detectedLanguageCode) || 'en';
+			languageCodes.find((c) => c === metadata.detectedLanguageCode) || 'auto';
 		const transcription = await runTranscription(
 			whisperBaseParams,
 			languageCode,
 			false,
 		);
+
+		transcription.metadata.detectedLanguageCode = metadata.detectedLanguageCode;
 		const translation =
 			languageCode === 'en'
 				? null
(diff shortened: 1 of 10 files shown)


scripts/purge-local-queues.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/zsh
+
+aws sqs purge-queue --queue-url http://localhost:4566/000000000000/transcription-service-task-queue-DEV.fifo --endpoint-url http://localhost:4566
+aws sqs purge-queue --queue-url http://localhost:4566/000000000000/transcription-service-output-queue-DEV --endpoint-url http://localhost:4566
+aws sqs purge-queue --queue-url http://localhost:4566/000000000000/giant-output-dead-letter-queue-DEV.fifo --endpoint-url http://localhost:4566
+aws sqs purge-queue --queue-url http://localhost:4566/000000000000/transcription-service-task-dead-letter-queue-DEV.fifo --endpoint-url http://localhost:4566
+aws sqs purge-queue --queue-url http://localhost:4566/000000000000/giant-output-queue-DEV.fifo --endpoint-url http://localhost:4566
```
philmcmahon, line 7: very useful!

scripts/purge-local-queues.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/zsh
```
philmcmahon, line 1: should be bash not zsh in case someone doesn't have zsh

scripts/delete-local-queues.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/zsh
```
philmcmahon, line 1: should be bash not zsh in case someone doesn't have zsh

philmcmahon (approved):
Looking forwards to seeing this working!

pipes

packages/worker/src/transcribe.ts
```diff
@@ -222,12 +222,14 @@ const transcribeAndTranslate = async (
...
 		const { metadata } = await runWhisper(whisperBaseParams, dlParams);
 		const languageCode =
-			languageCodes.find((c) => c === metadata.detectedLanguageCode) || 'en';
+			languageCodes.find((c) => c === metadata.detectedLanguageCode) || 'auto';
 		const transcription = await runTranscription(
 			whisperBaseParams,
 			languageCode,
 			false,
 		);
+
```
philmcmahon, line 231: could you add a comment here - something like `we only run language detection once, so need to override the detected language of future whisper runs`

PR #104, opened 2024-10-03
Add media download service (part 1)

philmcmahon:
## What does this change?
This pr introduces a new app to the transcription service - 'media download'. This service reads off a (new) SQS queue, parses the messages into a new `MediaDownloadJob` type, and then downloads the relevant url, then uploads the resulting file to the 'source media bucket' which is used by the transcription service, and sends an SQS message to the transcription worker to tell it to transcribe the file

To facilitate testing this service, I've also added a new endpoint to the transcription API, which pushes messages onto the queue based off the request body. You can test this endpoint by making a POST request to localhost:9103/api/transcribe-url with the following body:

```
{"url":"https://www.youtube.com/watch?v=upMOIAwWXmM","languageCode":"en","translationRequested":false}
```

Note - the infrastructure for the media download service doesn't exist yet, the project can only be run locally at the moment. 

## How to test
You'll need to manually install yt-dlp (https://github.com/yt-dlp/yt-dlp?tab=readme-ov-file#installation) - probably `pip install yt-dlp` is the easiest if you've got a suitable pythong environment set up. I haven't scripted this step because eventually we'll run yt-dlp using a docker container rather than isntalling it directly onto dev machines. This is how the existing whisper process works

If you run the transcription API, send the above request. If you get a 200 back, then run `npm run media-download::start` and you should see it read sqs/download the video/upload file to s3/send a transcription sqs message

## Still todo
 - UI for providing youtube url
 - Infra for media download service
 - Update output handler and export UI to support saving the original video to google drive


packages/cdk/lib/transcription-service.ts
```diff
@@ -462,6 +462,24 @@ export class TranscriptionService extends GuStack {
...
+		const mediaDownloadTaskQueue = new Queue(
+			this,
+			`${APP_NAME}-media-download-task-queue`,
+			{
+				fifo: true,
+				queueName: `${APP_NAME}-media-download-task-queue-${this.stage}.fifo`,
+				visibilityTimeout: Duration.seconds(30),
+				contentBasedDeduplication: true,
+				deadLetterQueue: {
+					queue: transcriptionDeadLetterQueue,
```
zekehuntergreen, line 475: Do we want transcription and download to share a dead letter queue? It seems like debugging might be easier if they're separate.

packages/backend-common/src/sqs.ts
```diff
@@ -87,12 +87,26 @@ export const generateOutputSignedUrlAndSendMessage = async (
...
+		translate: false,
 	};
-	return await sendMessage(client, queueUrl, JSON.stringify(job), s3Key);
+	const messageResult = await sendMessage(
+		client,
+		queueUrl,
+		JSON.stringify(job),
+		s3Key,
+	);
+	if (!isSqsFailure(messageResult) && translationRequested) {
```
zekehuntergreen, line 98: might want to log an error if messageResult is an sqs failure

packages/backend-common/src/sqs.ts
```diff
@@ -66,18 +66,18 @@ export const generateOutputSignedUrlAndSendMessage = async (
 	originalFilename: string,
 	inputSignedUrl: string,
 	languageCode: LanguageCode,
-	translate: boolean,
+	translationRequested: boolean,
```
zekehuntergreen, line 69: this is a better name 👍

packages/media-download/src/yt-dlp.ts
```diff
@@ -0,0 +1,35 @@
...
+	};
+};
+
+export const downloadMedia = async (
+	url: string,
+	destinationDirectoryPath: string,
+	id: string,
+) => {
+	const output =
+		await $`yt-dlp --write-info-json --no-clean-info-json --newline -o "${destinationDirectoryPath}/${id}" ${url}`;
```
zekehuntergreen, line 28: has yt-dlp been pretty consistent for youtube? I was noticing some flakiness using yt-dlp to download videos today.

scripts/setup.sh
```diff
@@ -45,7 +52,8 @@ DYNAMODB_ARN=$(aws --endpoint-url=http://localhost:4566 dynamodb create-table \
 
 echo "Created table, arn: ${DYNAMODB_ARN}"
 
-
+# media-downloader dependencies
+brew install ffmpeg phantomjs
```
zekehuntergreen, line 56: what about installing yt-dlp itself?

packages/media-download/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+		config.aws.localstackEndpoint,
+	);
+	const job = await getNextJob(
+		sqsClient,
+		config.app.mediaDownloadQueueUrl,
+		config.app.stage === 'DEV',
+	);
+	if (job) {
+		const metadata = await downloadMedia(job.url, '/tmp', job.id);
+		const key = await uploadToS3(
```
zekehuntergreen, line 92: should we remove the file from the file system once its uploaded?

packages/backend-common/src/sqs.ts
```diff
@@ -87,12 +87,26 @@ export const generateOutputSignedUrlAndSendMessage = async (
...
+		JSON.stringify(job),
+		s3Key,
+	);
+	if (!isSqsFailure(messageResult) && translationRequested) {
+		return await sendMessage(
+			client,
+			queueUrl,
+			JSON.stringify({ ...job, translate: true }),
+			s3Key,
+		);
```
zekehuntergreen, line 104: I'm not sure I understand this bit - we put two messages in the queue if the user has requested a translation? Doesn't the worker create the translation as its transcribing?

packages/cdk/lib/transcription-service.ts
philmcmahon: eek good spot! This was a copy paste failure, I agree they should be separate

packages/media-download/src/yt-dlp.ts
philmcmahon: as discussed, I've had some success with it. I'm hopeful that even if we end up having to use a different tool most of this code will still be useful

packages/media-download/src/index.ts
philmcmahon: At the moment I'm hoping we can run media-download on an ECS fargate task (like lurch ingest jobs) so the whole container will be deleted after the download/upload has finished

packages/backend-common/src/sqs.ts
philmcmahon: The way translation works at the moment is that we send two messages and the translation runs on one worker, the normal transcription on the other. My idea there was to speed up the transcription/translation - if we do it all on the same worker then it could take twice as long

scripts/setup.sh
philmcmahon: Good point. I've updated the pr description to address this, and now removed this brew command from setup.sh.

The way whisper works at the moment is that the worker runs a docker container with whisper installed rather than actually executing whisper directly. I think we could do the same thing for media-download. 

At the moment there's no guidance on installing yt-dlp in this PR, pending the next PR with various dockerfiles and other infrastructure bits in it. I could add a Pipfile (pipenv is my fav new python thing since you showed it to me) but given that it will just be for a short while before we swap out running yt-dlp directly for docker maybe it's fine as is

philmcmahon:
@zekehuntergreen thanks v much for the review. I ran into an issue related to zx deprecating commonjs support and decided (again I think!) to swap it out and use the basic node subprocess command instead - that change is here https://github.com/guardian/transcription-service/pull/104/commits/322acafb1d1d149858c1e153f9fe45fc2c949f47

packages/backend-common/src/sqs.ts
zekehuntergreen: Ah ok thanks, I hadn't realized this.

scripts/setup.sh
zekehuntergreen: Ok, happy to wait for yt-dlp on docker

packages/backend-common/src/process.ts
```diff
@@ -0,0 +1,48 @@
...
+			stderr.push(e.toString());
+		});
+
+		cp.on('close', (code) => {
+			const result = {
+				stdout: stdout.join(''),
+				stderr: stderr.join(''),
+				code: code || undefined,
+			};
+			logger.info('Ignoring stdout to avoid logging sensitive data');
```
zekehuntergreen, line 36: Is it important not to log stdout when running yt-dlp?

zekehuntergreen (approved):
download-percentage

PR #110, opened 2024-11-12
Remove beta tag

philmcmahon:
## What does this change?
Given expected increase in the tool usage and relative maturity, let's drop the beta tag maybe?


hoyla (approved):
👍

PR #113, opened 2024-12-20
Bump to latest gucdk version 61.0.2

philmcmahon:
## What does this change?
The transcription service was previously relying on a branch of GuCDK, this PR sets the version to release version 61.0.2, in preparation for doing this https://github.com/guardian/lurch/pull/751 to the transcription service.

cpuArchitecture is no longer a parameter of the ecs task pattern - it just always uses ARM64


srbd (approved):
nice

PR #114, opened 2025-01-06
 Bump CDK to support having readonlyRootFilesystem set to true #751 

philmcmahon:
## What does this change?
Similar to https://github.com/guardian/lurch/pull/751, this PR disables write access to the root file system for ecs containers used in this project (just the media download service at the moment). This is achieved by bumping the version of GuCDK that the project uses to pull in https://github.com/guardian/cdk/pull/2545

With this in place, we then need to make sure that all file write operations happen inside a separate mounted volume - in this case mounted at /media-download

## How to test
Tested on CODE

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,6 +64,41 @@
 			"dev": true,
 			"license": "Apache-2.0"
 		},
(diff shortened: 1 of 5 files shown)


srbd (approved):
cool

PR #117, opened 2025-01-10
Improve url form validation

philmcmahon:
## What does this change?
It would be helpful to perform some client side validation on urls being added to the upload form so that we can, e.g. prevent people submitting invalid urls, or urls to youtube search results (as yt-dlp will then try download everything on the page). 

This PR introduces this, by breaking up the existing 'textarea' into individual input elements, as suggested by @zekehuntergreen ages ago. 

Before:
Screenshot 2025-01-10 at 16 43 30

After:

Screenshot 2025-01-10 at 16 44 24

## How to test
You can run just the API and the client locally to test this - though hopefully screenshots make it fairly clear what's going on. I've not tested on CODE yet as I'm waiting for https://github.com/guardian/transcription-service/pull/115 to go out.

### DIFF
--- a/packages/client/src/components/UploadForm.tsx
+++ b/packages/client/src/components/UploadForm.tsx
@@ -121,26 +121,26 @@ const updateFileStatus = (
 	};
 };
 
-const checkUrlInputValid = (input: MediaUrlInput): MediaUrlInput => {
+const checkUrlValid = (url_input: string): MediaUrlInput => {
 	try {
-		if (input.value === '') {
-			return { value: input.value, status: 'empty' };
+		if (url_input === '') {
+			return { value: url_input, status: 'empty' };
 		}
-		const url = new URL(input.value);
+		const url = new URL(url_input);
 		// we don't want people providing search results pages as yt-dlp will try and fetch every video
 		if (
 			url.pathname.includes('results') &&
 			url.search.includes('search_query')
 		) {
 			return {
-				value: input.value,
+				value: url_input,
 				reason: 'URL is a link to search results. Please link to a video page',
 				status: 'invalid',
 			};
 		}
-		return { value: input.value, status: 'valid' };
+		return { value: url_input, status: 'valid' };
 	} catch {
-		return { ...input, reason: 'Invalid URL', status: 'invalid' };
+		return { value: url_input, reason: 'Invalid URL', status: 'invalid' };
 	}
 };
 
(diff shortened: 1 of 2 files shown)


hoyla:
In the "after" view it wasn't immediately clear to me which problem pertained to which URL, and I think it might be worse for red/green colourblind users. Maybe put a divider line between each URL field, so that any problem reporting appears adjacent to the URL it pertains to, with a line between that text and the following URL?

hoyla:
Oh, while you're tweaking this text can you use "URL" rather than "Url"? In both the widget label and on the Add Url button

packages/client/src/components/UploadForm.tsx
```diff
@@ -108,12 +121,26 @@ const updateFileStatus = (
 	};
 };
 
-const checkUrlValid = (url: string) => {
+const checkUrlInputValid = (input: MediaUrlInput): MediaUrlInput => {
```
zekehuntergreen, line 124: does this parameter need to change? it seems like the function is simpler if it continues to accept the url string to validate.

packages/client/src/components/UploadForm.tsx
```diff
@@ -285,19 +318,45 @@ export const UploadForm = () => {
...
+												setMediaUrlInputs(
+													mediaUrlInputs.map((input, i) =>
+														i === index
+															? checkUrlInputValid({
+																	...input,
+																	value: e.target.value,
+																})
+															: input,
+													),
+												);
```
zekehuntergreen, line 345: Seems like this could be a bit simpler if we called a helper function that indexed into `mediaUrlInputs` rather than mapping
```typescript
	const setMediaUrlInput = (index: number, value: string) => {
		const newInputs = [...mediaUrlInputs];
		newInputs[index] = checkUrlInputValid(value);
		setMediaUrlInputs(newInputs);
	};
```
I think this is just a matter of personal preference though. Your way we avoid allocating a new array.

zekehuntergreen (approved):
looks good! just a few non-blocking comments

packages/client/src/components/UploadForm.tsx
philmcmahon: I decided to do it inline but otherwise think this is a sensible improvement. I tend to try and keep things functional as if writing scala but it doesn't make sense given that I'm mutating stuff all over the place elsewhere!

PR #118, opened 2025-01-14
Store language code in database

philmcmahon:
## What does this change?
I'm currently experimenting with https://github.com/m-bain/whisperX. WhisperX appears to [support fewer languages](https://github.com/m-bain/whisperX?tab=readme-ov-file#other-languages) than whisper.cpp. To understand the impact of moving to whisperX this PR sets the transcription service to record the language code of the transcription in the dynamo database.

I made the new field optional in the zod type to avoid breaking existing transcriptions. In a few weeks we could make it compulsory.

## How to test
Tested on CODE

### DIFF
--- a/packages/backend-common/src/dynamodb.ts
+++ b/packages/backend-common/src/dynamodb.ts
@@ -5,9 +5,8 @@ import {
 	GetCommand,
 } from '@aws-sdk/lib-dynamodb';
 
-import { z } from 'zod';
 import { logger } from '@guardian/transcription-service-backend-common';
-import { ExportStatuses } from '@guardian/transcription-service-common';
+import { TranscriptionDynamoItem } from '@guardian/transcription-service-common';
 
 export const getDynamoClient = (
 	region: string,
@@ -24,27 +23,6 @@ export const getDynamoClient = (
 	const client = new DynamoDBClient(clientConfig);
 	return DynamoDBDocumentClient.from(client);
 };
-
-export const TranscriptKeys = z.object({
-	srt: z.string(),
-	text: z.string(),
-	json: z.string(),
-});
-
-export type TranscriptKeys = z.infer<typeof TranscriptKeys>;
-
-export const TranscriptionDynamoItem = z.object({
-	id: z.string(),
-	originalFilename: z.string(),
-	transcriptKeys: TranscriptKeys,
-	userEmail: z.string(),
-	completedAt: z.optional(z.string()), // dynamodb can't handle dates so we need to use an ISO date
-	isTranslation: z.boolean(),
-	exportStatuses: z.optional(ExportStatuses),
-	languageCode: z.optional(z.string()),
-});
-
-export type TranscriptionDynamoItem = z.infer<typeof TranscriptionDynamoItem>;
 
 export const writeTranscriptionItem = async (
 	client: DynamoDBDocumentClient,
--- a/packages/backend-common/src/dynamodb.ts
+++ b/packages/backend-common/src/dynamodb.ts
@@ -41,6 +41,7 @@ export const TranscriptionDynamoItem = z.object({
 	completedAt: z.optional(z.string()), // dynamodb can't handle dates so we need to use an ISO date
 	isTranslation: z.boolean(),
 	exportStatuses: z.optional(ExportStatuses),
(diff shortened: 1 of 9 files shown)


packages/backend-common/src/dynamodb.ts
```diff
@@ -39,6 +39,7 @@ export const TranscriptionDynamoItem = z.object({
 	userEmail: z.string(),
 	completedAt: z.optional(z.string()), // dynamodb can't handle dates so we need to use an ISO date
 	isTranslation: z.boolean(),
+	languageCode: z.optional(z.string()),
```
zekehuntergreen, line 42: Is this an iso country code? If so this type could be more narrow if we use our [`LanguageCode`](https://github.com/guardian/transcription-service/blob/71f91751872589d9c02e11112239e1bd1de6d80c/packages/common/src/languages.ts#L111) type

packages/common/src/types.ts
```diff
@@ -1,5 +1,5 @@
 import { z } from 'zod';
-import { languageCodeToLanguage } from './languages';
+import { LanguageCode, languageCodeToLanguage } from './languages';
```
zekehuntergreen, line 2: I think your zod enum means we can get rid of `zodLanguageCode` just below which does the same thing (and also `getKeys`). Your solution of making a zod enum from the array of country codes is way simpler than what I did!

packages/common/src/languages.ts
```diff
@@ -108,11 +109,12 @@ export const languageCodeToLanguage = Object.freeze({
...
 // is an array of a union of string literals rather than an array of strings.
 // https://www.charpeni.com/blog/properly-type-object-keys-and-object-entries
-export const languageCodes = Object.keys(
-	languageCodeToLanguage,
-) as unknown as LanguageCode[];
+export const languageCodes = Object.keys(languageCodeToLanguage) as [
+	keyof LanguageCodeToLanguage,
+];
+
+export const LanguageCode = z.enum(languageCodes);
```
zekehuntergreen, line 119: I think zod objects are camelCase cause they're just consts? although we're not following this convention everywhere
```suggestion
export const languageCode = z.enum(languageCodes);
```

packages/output-handler/src/index.ts
```diff
@@ -63,6 +64,14 @@ const handleTranscriptionSuccess = async (
 	metrics: MetricsService,
 	sourceMediaDownloadUrl: string,
 ) => {
+	const languageCode = LanguageCode.safeParse(transcriptionOutput.languageCode);
```
zekehuntergreen, line 67: In [`processMessage`](https://github.com/guardian/transcription-service/blob/41b54d899416fe667deb164783a697aab0e7420a/packages/output-handler/src/index.ts#L166) we're parsing the sqs event object before calling this function. It might be cleaner to change [`TranscriptionOutputSuccess`](https://github.com/guardian/transcription-service/blob/41b54d899416fe667deb164783a697aab0e7420a/packages/common/src/types.ts#L75) so that its languageCode is an enum than to do a second round of parsing here.
Although I'm now realising that this would be more of a refactor of our types and zod parsers rather than the original goal of just storing language! So feel free to ignore this or just put in a TODO

zekehuntergreen (approved):

packages/common/src/languages.ts
philmcmahon: I don't think we're very consistent with this. I think that having camelCase for the zod object and TitleCase for the type probably makes sense as sometimes you need to specifically reference the zod object rather than the type. On the other hand, it's quite nice not having it all as title case and just being able to do : Type or Type.safeParse

packages/output-handler/src/index.ts
philmcmahon: This opened up a rabbit hole, I've ended up refactoring the language types quite a bit in https://github.com/guardian/transcription-service/pull/118/commits/24780a33ae7ec9ec9b6f867c5a93ae5e483a13e4 

The idea is - going in the type is either a code or auto, coming out either 'unknown' or a valid language code

zekehuntergreen: thanks for the refactor! I think your input and output types are way cleaner than what we were doing before

zekehuntergreen (approved):
looks good!

PR #121, opened 2025-01-20
Create media export function parameter in cdk

philmcmahon:
## What does this change?

When deploying https://github.com/guardian/transcription-service/pull/115 it didn't work because I left a whitespace character at the end of the mediaExportFunctionName parameter. Since this param just refers to something created in cdk we should create it in cdk.

## How to test
Tested on CODE - need to delete the existing param before deploying this change


philmcmahon:
This was inadvertently merged as part of https://github.com/guardian/transcription-service/pull/123

PR #123, opened 2025-01-23
Add whisperx support (including diarization)

philmcmahon:
## What does this change?
This adds support for using https://github.com/m-bain/whisperX on the transcription workers.

The main benefit of whisperx is that it supports diarization, or speaker recognition. This is a frequently requested feature of the transcription tool.

The downside of whisperx is that it needs to be run on a GPU instance. So far I've been running it on a g5.xlarge instance. These instances are roughly 2x the price of the c7g.4xlarge instances we've been using to run whisper.cpp on. I think we might get some savings moving to a g4dn.2xlarge instance, but performance testing would be required for that. See [here](https://instances.vantage.sh/?filter=g4dn.2xlarge|g5.xlarge|c7g.4xlarge&compare_on=true&selected=g4dn.2xlarge,g5.xlarge,c7g.4xlarge) for the cost figures. I think that the cost increase may be a little more than 2x, as I imagine that GPU instances are less available on the spot market.

**Update:** Performance is actually better on a g4dn.2xlarge than on the g5.xlarge instance, and 25% cheaper at $0.75/hr - so only a 30% increase over our existing instances (the spot issue probably remains)

As well as cost, the other downside of whisper x is that getting it running (and using the gpu) is a right ole faff. To get it working I needed to:
 - Create a new amigo base image based off the [aws deep learning ami](https://docs.aws.amazon.com/dlami/latest/devguide/overview-base.html) - see https://github.com/guardian/amigo/pull/1604
 - Create a new whisperx amigo role - https://github.com/guardian/amigo/pull/1606
 - Pre-cache all the various whisperx models into the AMI using the python script included in this pr, and https://github.com/guardian/amigo/pull/1607

I'm not yet certain that whisperx has sufficient performance improvements such that we'll want to use it for all transcripts - it may be we only want to use it where the user has requested diarization. With that in mind, I've set whisperx up as a totally separate pipeline, with its own SQS queue and autoscaling group. Currently the use of whisperx (and the gpu instances) is controlled by a parameter store parameter, and is enabled for DEV/CODE and disabled for PROD. 

I decided to install whisperx onto the base AMI rather than using a containerised version as we are doing with whisper.cpp. Whilst you can fairly easily allow docker containers to make use of the GPU, I didn't see much benefit of adding an extra layer of virtualisation. This change has lead to some awkwardness as 'useWhisperX' also means "don't use whisper.cpp container for ffmpeg or transcription' - it will be good if whisperX proves superioir across the board so that we can remove all this conditional logic.

## Reviewing this PR
The most interesting changes are to the worker app - everything else is just wiring really. Even the worker app is now just running whisperx instead of whispercpp. 

## How to test
I've tested this on CODE using an AMI baked on amigo CODE - I'll need to get https://github.com/guardian/amigo/pull/1607 merged before we can get a PROD AMI. 

Happy users thanks to the new diarization feature!

Current performance tests:
 - 1h7 minute audio https://www.youtube.com/watch?v=uKK1RX4Ml3I
 - whisper.cpp - 33mins
 - whisperx with diarization - 20mins

I don't think diarization adds much to the time.

I think the performance boost would be much higher if we could pre-warm the gpu instance somehow, but need to do more research there.


philmcmahon:
An update - the offline issue is fixed in https://github.com/m-bain/whisperX/pull/1021 and https://github.com/guardian/transcription-service/pull/123/commits/163505e691bfa1ddabeac25f9ea7f1c76e3f529a

Some more performance measurements:



 - 1h7 minute audio https://www.youtube.com/watch?v=uKK1RX4Ml3I
 - whisperx with diarization, g5.xlarge, phoneme alignment disabled - 16mins

package.json
```diff
@@ -7,18 +7,18 @@
...
 		"worker-capacity-manager::build": "npm run build --workspace worker-capacity-manager",
-		"worker-capacity-manager::start": "AWS_REGION=eu-west-1 STAGE=DEV npm run start --workspace worker-capacity-manager",
+		"worker-capacity-manager::start": "APP=worker-capacity-manager AWS_REGION=eu-west-1 STAGE=DEV npm run start --workspace worker-capacity-manager",
 		"media-download::build": "npm run build --workspace media-download",
-		"media-download::start": "AWS_REGION=eu-west-1 STAGE=DEV npm run start --workspace media-download",
+		"media-download::start": "APP=media-download AWS_REGION=eu-west-1 STAGE=DEV npm run start --workspace media-download",
 		"worker::build": "npm run build --workspace worker; npm run build --workspace worker",
 		"worker::package": "npm run package --workspace worker",
-		"worker::start": "AWS_REGION=eu-west-1 STAGE=DEV npm run start --workspace worker",
+		"worker::start": "APP=transcription-service-gpu-worker AWS_REGION=eu-west-1 STAGE=DEV npm run start --workspace worker",
```
philmcmahon, line 21: todo: add worker-gpu::start, worker-cpu::start

packages/cdk/lib/transcription-service.ts
```diff
@@ -616,18 +725,31 @@ export class TranscriptionService extends GuStack {
...
-			name: `${mediaDownloadApp}-volume`,
+		const downloadVolume = {
+			name: `${mediaDownloadApp}-download-volume`,
 		};
-		mediaDownloadTask.taskDefinition.addVolume(volume);
+		const tempVolume = {
+			name: `${mediaDownloadApp}-temp-volume`,
+		};
+		mediaDownloadTask.taskDefinition.addVolume(downloadVolume);
+		mediaDownloadTask.taskDefinition.addVolume(tempVolume);
```
philmcmahon, line 742: todo - pull this out into a different PR as it's unrelated whisperX

packages/worker/src/transcribe.ts
```diff
@@ -267,11 +297,58 @@ const whisperParams = (
...
+			'--model_cache_only',
+			'True',
+			'--output_dir',
+			path.parse(wavPath).dir,
+			wavPath,
+		]);
+		const metadata = extractWhisperXStderrData(result.stderr);
+		logger.info('Whisper finished successfully', metadata);
+		return {
+			fileName: `${fileName}`,
```
zekehuntergreen, line 333:
```suggestion
			fileName,
```

packages/worker/src/transcribe.ts
```diff
@@ -172,27 +176,45 @@ const runTranscription = async (
...
 
+const getLanguageCode = async (
+	whisperBaseParams: WhisperBaseParams,
+	whisperX: boolean,
+): Promise<LanguageCode> => {
+	if (whisperX) {
+		return Promise.resolve('auto');
+	}
+	const dlParams = whisperParams(true, whisperBaseParams.wavPath);
+	const { metadata } = await runWhisper(whisperBaseParams, dlParams);
```
zekehuntergreen, line 187: this was the case before your change, but are we running whisper twice (or thrice if we're translating)? once here and once in `runTranscription`?

zekehuntergreen: might be worth adding a comment to explain

packages/worker/src/transcribe.ts
```diff
@@ -172,27 +176,45 @@ const runTranscription = async (
 	}
 };
 
+const getLanguageCode = async (
+	whisperBaseParams: WhisperBaseParams,
+	whisperX: boolean,
+): Promise<LanguageCode> => {
+	if (whisperX) {
+		return Promise.resolve('auto');
```
zekehuntergreen, line 184: does whisperx not take a language arg?

scripts/download_whisperx_models.py
```diff
@@ -0,0 +1,136 @@
+import torchaudio
```
zekehuntergreen, line 1: might be worth adding a comment linking to amigo role where this is used

zekehuntergreen (approved):
Looks good 👏 

Thanks for the walkthrough

package.json
philmcmahon: done

packages/cdk/lib/transcription-service.ts
philmcmahon: done

package.json
```diff
@@ -7,18 +7,18 @@
 		"prettier:check": "prettier . --check",
 		"prettier:fix": "prettier . --write",
 		"api::build": "npm run build --workspace api",
-		"api::start": "AWS_REGION=eu-west-1 STAGE=DEV npm run start --workspace api",
+		"api::start": "APP=api AWS_REGION=eu-west-1 STAGE=DEV npm run start --workspace api",
```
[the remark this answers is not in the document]
philmcmahon: I went for APP because we are already tagging the instances with APP - it's just the start script that needs changing, whereas with a new variable I'd need to tag the instances accordingly

packages/cdk/lib/transcription-service.ts
```diff
@@ -365,29 +374,51 @@ export class TranscriptionService extends GuStack {
 			Port.tcp(443),
 		);
 
+		const launchTemplateProps = {
+			machineImage: MachineImage.genericLinux({
+				'eu-west-1': workerAmi.valueAsString,
+			}),
+			instanceType: InstanceType.of(InstanceClass.C7G, InstanceSize.XLARGE4),
```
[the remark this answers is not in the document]
philmcmahon: You're right, I was just being lazy. Resolved in https://github.com/guardian/transcription-service/pull/123/commits/c34debcec9a2f0cce3b94a5e555980c5ad3e493b

packages/worker/src/transcribe.ts
philmcmahon: whisperx takes so long to start up I didn't think there was any point running a pre-pass of language detection and decided to just detect it every time

PR #124, opened 2025-01-27
Stop auto deleting containers tagged with 'main

philmcmahon:
## What does this change?

We currently have a lifecycle rule for our ECR repositories that deletes all but the most recent 5 containers. This is causing problems on PROD, as if a development branch has more than 5 builds then the container tagged with `main` will end up being deleted.

This PR protects containers tagged with `main` from the lifecycle rule.

I've also bumped the number of containers from 5 to 10, 5 seemed a bit small because if you have 2 devs working on 2 branches then they could very quickly end up losing containers needed to test stuff.

## How to test
Deploy this, kick off a few builds, verify that `main` container isn't deleted

### DIFF
--- a/packages/cdk/lib/repository.ts
+++ b/packages/cdk/lib/repository.ts
@@ -58,7 +58,7 @@ export class TranscriptionServiceRepository extends GuStack {
 						rulePriority: 1,
 					},
 					{
-						maxImageCount: 10,
+						maxImageCount: 20,
 						rulePriority: 2,
 					},
 				],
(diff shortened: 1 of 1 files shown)


packages/cdk/lib/repository.ts
```diff
@@ -37,7 +37,12 @@ export class TranscriptionServiceRepository extends GuStack {
 				repositoryName: `transcription-service`,
 				lifecycleRules: [
 					{
-						maxImageCount: 5,
+						tagPrefixList: ['main'],
+						maxImageCount: 20,
+						rulePriority: 1,
+					},
+					{
+						maxImageCount: 10,
```
philmcmahon, line 45: add rulePriority:2

zekehuntergreen (approved):

PR #125, opened 2025-01-28
Use different ID, output urls for translation and transcript

philmcmahon:
## What does this change?
There's a bug in the transcription tool at the moment where both the transcription and the translation end up with the same id, resulting in whichever one finishes last overwriting the other! This fixes that.

### DIFF
--- a/packages/backend-common/src/sqs.ts
+++ b/packages/backend-common/src/sqs.ts
@@ -71,7 +71,6 @@ export const generateOutputSignedUrlAndSendMessage = async (
 	translationRequested: boolean,
 	diarizationRequested: boolean,
 ): Promise<SendResult> => {
-	const translationId = `${s3Key}-translation`;
 	const signedUrls = await generateOutputSignedUrls(
 		s3Key,
 		config.aws.region,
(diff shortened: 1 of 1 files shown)


packages/backend-common/src/sqs.ts
```diff
@@ -105,7 +111,12 @@ export const generateOutputSignedUrlAndSendMessage = async (
 		return await sendMessage(
 			client,
 			queueUrl,
-			JSON.stringify({ ...job, translate: true }),
+			JSON.stringify({
+				...job,
+				id: `${s3Key}-translation`,
```
zekehuntergreen, line 116: while we're at it it feels weird to define this here and in [`generateOutputSignedUrls`](https://github.com/guardian/transcription-service/pull/125/files#diff-7a6f98c31cdd871b93d5d3354f444d5d94916552fd05005856cdf0d460239415R318). Might be cleaner to pass filename to that function?

packages/backend-common/src/sqs.ts
```diff
@@ -74,12 +74,18 @@ export const generateOutputSignedUrlAndSendMessage = async (
 		outputBucket,
 		userEmail,
 		7,
-		translationRequested,
+		false,
+	);
+	const signedUrlsTranslation = await generateOutputSignedUrls(
```
zekehuntergreen, line 79: could go inside `if (!isSqsFailure(messageResult) && translationRequested)`

zekehuntergreen (approved):
well done spotting this!

packages/backend-common/src/sqs.ts
philmcmahon: good shout, definitely want to minimise the number of places we're changing the id to reduce problems exactly like this one

PR #129, opened 2025-01-31
Add /tmp volume to media download task container

philmcmahon:
## What does this change?
I've seen occasionaly failures in the media download service because yt-dlp yt-dlp needs to be able to write to the /tmp directory. In https://github.com/guardian/transcription-service/pull/114 I made the media export filesystem read only, allowing writes to a /media-download mounted volume. This PR adds an extra volume to support writing to /tmp.

## How to test
I've tested this on CODE. As it doesn't happen every time it's hard to test properly but I think it's a fairly harmless change.

### DIFF
--- a/packages/cdk/lib/transcription-service.ts
+++ b/packages/cdk/lib/transcription-service.ts
@@ -619,13 +619,22 @@ export class TranscriptionService extends GuStack {
 			],
 		});
 
-		const volume = {
-			name: `${mediaDownloadApp}-volume`,
+		const downloadVolume = {
+			name: `${mediaDownloadApp}-download-volume`,
 		};
-		mediaDownloadTask.taskDefinition.addVolume(volume);
+		const tempVolume = {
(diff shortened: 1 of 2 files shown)


packages/cdk/lib/transcription-service.ts
```diff
@@ -619,15 +619,24 @@ export class TranscriptionService extends GuStack {
 			],
 		});
 
-		const volume = {
-			name: `${mediaDownloadApp}-volume`,
+		const downloadVolume = {
+			name: `${mediaDownloadApp}-download-volume`,
 		};
-		mediaDownloadTask.taskDefinition.addVolume(volume);
+		const tempVolume = {
```
zekehuntergreen, line 625: might be worth a comment explaining why this is needed

zekehuntergreen (approved):

packages/cdk/lib/transcription-service.ts
philmcmahon: good shout

PR #130, opened 2025-01-31
Add CI step for models script

philmcmahon:
## What does this change?
To support this change in AMIgo https://github.com/guardian/amigo/pull/1607 to download whisperx models, this PR adds a CI step to add a new riffraff project which deploys the download-models script to a bucket in the deploy tools account where AMIgo can access it.

See this discussion for the motivation to publish to S3 via riffraff https://github.com/guardian/amigo/pull/1607#discussion_r1933535778

With this change, if we want to make a change to the download model script, the steps are:

 - push to feature branch
 - use riffraff to deploy the investigations::transcription-service-whisperx-model-fetch project to CODE/PROD
 - If we deployed to CODE, update the 'stage' parameter set on the AMIgo recipe used for the whisperx workers to point at CODE
 - Bake an AMI - it will fetch the script we deployed with riffraff

## How to test
I've tested this using AMIgo CODE


zekehuntergreen (approved):

whisperx-model-fetch/riff-raff.yaml
```diff
@@ -0,0 +1,17 @@
+allowedStages:
+  - PROD
```
zekehuntergreen, line 2: no CODE?

philmcmahon: oops!

PR #135, opened 2025-02-03
Use original file name when downloading source media

philmcmahon:
## What does this change?
Currently, when you get an output email from the transcription service and click the link to download source media, you get a file called UUID, which isn't very user friendly. This PR sets the filename to the original filename using the 'Content-Disposition' header https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

## How to test
Tested on CODE

### DIFF
--- a/packages/backend-common/src/s3.ts
+++ b/packages/backend-common/src/s3.ts
@@ -55,6 +55,12 @@ export const getSignedUploadUrl = (
 	);
 };
 
+const sanitizeFilename = (filename: string) => {
+	let sanitized = filename.replace(/[^a-zA-Z0-9._-]/g, '_');
+	sanitized = sanitized.substring(0, 255);
+	return sanitized;
+};
+
 export const getSignedDownloadUrl = async (
 	region: string,
 	bucket: string,
@@ -63,7 +69,7 @@ export const getSignedDownloadUrl = async (
 	overrideFilename?: string,
 ) => {
 	const responseContentDisposition = overrideFilename
-		? `attachment; filename="${overrideFilename}"`
+		? `attachment; filename="${sanitizeFilename(overrideFilename)}"`
 		: undefined;
 	return await getSignedUrlSdk(
 		getS3Client(region),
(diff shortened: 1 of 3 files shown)


hoyla:
🙏

packages/backend-common/src/s3.ts
```diff
@@ -60,15 +60,21 @@ export const getSignedDownloadUrl = async (
 	bucket: string,
 	key: string,
 	expiresIn: number,
-) =>
-	await getSignedUrlSdk(
+	overrideFilename?: string,
+) => {
+	const responseContentDisposition = overrideFilename
+		? `attachment; filename="${overrideFilename}"`
```
zekehuntergreen, line 66: It might be worth sanitising this string since its provided by the user? 
I suppose because of file access controls, users could only do a header injection attack on themselves or the dev who downloads the file, but might be worth locking down anyway

packages/output-handler/src/index.ts
```diff
@@ -181,7 +182,15 @@ const processMessage = async (event: unknown) => {
 				sourceMediaDownloadUrl,
 			);
 		} else if (transcriptionOutputIsTranscriptionFailure(transcriptionOutput)) {
-			logger.info('handling transcription failure');
+			logger.info(
+				`Handling transcription failure. Transcription output: ${JSON.stringify(transcriptionOutput)}`,
+			);
+			const sourceMediaDownloadUrl = await getSignedDownloadUrl(
```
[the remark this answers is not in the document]
philmcmahon: Ah yes good point. It's only 'media download failure' that doesn't have an originalFilename - and I've realised we're not really handling that case at all! Something for a future pr

packages/backend-common/src/s3.ts
philmcmahon: yes great shout, added in https://github.com/guardian/transcription-service/pull/135/commits/65d3a5547227512dbca67a7edacd0b03a79e1b34

zekehuntergreen (approved):
looks good!

name

PR #136, opened 2025-02-05
Calculate duration of input files in order to decide which engine to use

philmcmahon:
## What does this change?
Currently we have two available 'engines' for transcription:
 - whisperx - supports diarization, is very fast once it gets started, better output formatting, very slow startup (~10mins)
 - whisper.cpp - fast(er) startup, less pretty output formatting, no diarization

It seems a shame to wait to spin up whisperx for a file that is only a few minutes long. This PR gets the API to make a decision on which engine to use based on the duration of the file - any files under 10 minutes long where the user hasn't requested diarization will be sent to whisper.cpp, everything else goes to whisperX

To get ffprobe running on lambda I had to create a lambda layer - this is basically just a zip file that gets added to the disk of the lambda vm. 

## How to test
Tested on CODE

### DIFF
--- a/packages/backend-common/src/ffmpeg.ts
+++ b/packages/backend-common/src/ffmpeg.ts
@@ -1,0 +1,19 @@
+import { exec } from 'node:child_process';
+import { logger } from '@guardian/transcription-service-backend-common';
+import { promisify } from 'node:util';
+
+const execPromise = promisify(exec);
+
+export const getFileDuration = async (filePath: string): Promise<number> => {
+	const command = `ffprobe -i ${filePath} -show_entries format=duration -v quiet -of csv="p=0"`;
+	try {
+		const { stdout, stderr } = await execPromise(command);
+		if (stderr) {
+			logger.error(` ffprobe stderr: `, stderr);
+		}
+		return parseFloat(stdout);
+	} catch (error) {
+		logger.error(`Error during ffprobe file duration detection`, error);
+		throw error;
+	}
+};
--- a/packages/backend-common/src/ffmpeg.ts
+++ b/packages/backend-common/src/ffmpeg.ts
@@ -4,16 +4,18 @@ import { promisify } from 'node:util';
 
 const execPromise = promisify(exec);
 
-export const getFileDuration = async (filePath: string): Promise<number> => {
+export const getFileDuration = async (
+	filePath: string,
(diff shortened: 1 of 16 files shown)


packages/backend-common/src/ffmpeg.ts
```diff
@@ -0,0 +1,19 @@
...
+	const command = `ffprobe -i ${filePath} -show_entries format=duration -v quiet -of csv="p=0"`;
+	try {
+		const { stdout, stderr } = await execPromise(command);
+		if (stderr) {
+			logger.error(` ffprobe stderr: `, stderr);
+		}
+		return parseFloat(stdout);
+	} catch (error) {
+		logger.error(`Error during ffprobe file duration detection`, error);
+		throw error;
```
zekehuntergreen, line 17: will we return a 500 if we can't find file duration here? might be better to default to whisperx

packages/common/src/types.ts
```diff
@@ -69,6 +69,8 @@ export const TranscriptionJob = z.object({
 	engine: z.nativeEnum(TranscriptionEngine),
 	// we can get rid of this when we switch to using a zip
 	translationOutputBucketUrls: z.optional(OutputBucketUrls),
+	// this is optional because giant currently doesn't know file duration
```
zekehuntergreen, line 72: do ffmpeg or the whisper models output file duration? Or would it be worth having the worker calculate duration with ffprobe? That would unify the way we're finding duration and make it so Giant isn't a special case.

packages/backend-common/src/ffmpeg.ts
philmcmahon: good shout, implemented in https://github.com/guardian/transcription-service/pull/136/commits/5d7709a195510dd82070c4b929c33d9578d7fa95

packages/common/src/types.ts
philmcmahon: Good point. I was more just shoving this in in case useful for logging - the only place it was used was in setting a 'duration' field in the transcription failure message, which isn't really necessary. So I've removed this field in https://github.com/guardian/transcription-service/pull/136/commits/244a6d2d5dbf09b189c44363abd5643a972e1a6f

zekehuntergreen (approved):

PR #138, opened 2025-02-07
Handle media download failures 

philmcmahon:
## What does this change?
I spotted that the media download serivice doesn't actually currently have permission to write to the transcription output queue, so if a media download fails it can't tell the user.

This fixes that, and adds a step to the output handler to send an email to the user explaining what might have happened. It looks like this:

Screenshot 2025-02-07 at 11 35 21

## How to test
Tested on CODE

### DIFF
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -73,10 +73,13 @@ export const TranscriptionJob = z.object({
 
 export type TranscriptionJob = z.infer<typeof TranscriptionJob>;
 
-const TranscriptionOutputBase = z.object({
+const OutputBase = z.object({
 	id: z.string(),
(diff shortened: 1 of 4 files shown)


packages/common/src/types.ts
```diff
@@ -88,7 +91,7 @@ export const TranscriptionOutputSuccess = TranscriptionOutputBase.extend({
 	translationOutputBucketKeys: z.optional(OutputBucketKeys),
 });
 
-export const MediaDownloadFailure = z.object({
+export const MediaDownloadFailure = OutputBase.extend({
```
zekehuntergreen, line 94: can get rid of `id`

zekehuntergreen (approved):
pipes

PR #139, opened 2025-02-21
Support transcripts with apostrophes in the filename

philmcmahon:
## What does this change?
I thought that the transcription service was broken for cyrillic, but it was actually broken for all files with an apostrophe in the filename. This resolves that by escaping apostrophes in the call to the google drive API

## How to test
I've tested this locally, think it's good to go


srbd (approved):
very good escaping

PR #148, opened 2025-04-01
Worker capacity manager finds ASG scaling limit

zekehuntergreen:
## What does this change?
Following #147 change the worker capacity manager so that it finds the scaling limits of each ASG rather than using hardcoded prod value

- [x] test in code


dominickendrick (approved):
Looks good

PR #152, opened 2025-07-04
Support and enable IMDSV2 for workers

philmcmahon:
## What does this change?
Enforcing IMDSv2 is a recommended security improvement. It's enforced by default for normal gu cdk apps but our ASG for the workers doesn't use the gucdk constructs, so need to do it manually.

We make a couple of calls in the workers to the metadata service, I've modified these to correctly fetch the token. 

## How to test
Needs testing on CODE, steps would be:
 - deploy to CODE but clouformation only so that IMDSV2 is blocked but the app code isn't updated
 - Run a transcription job
 - login and look out for 'Error during spot termination check' logs - hopefully there will be some
 - deploy the whole thing to CODE
 - do a transcription, make sure that it works

### DIFF
--- a/packages/worker/src/spot-termination.ts
+++ b/packages/worker/src/spot-termination.ts
@@ -52,7 +52,7 @@ export const checkSpotInterrupt = async (
 				return;
 			}
 		} else {
-			logger.info(
+			logger.error(
 				`Non-200 response from ${url}: ${result.status} ${result.statusText}`,
 			);
 		}


packages/cdk/lib/transcription-service.ts
```diff
@@ -454,6 +454,7 @@ export class TranscriptionService extends GuStack {
 					'eu-west-1': gpuWorkerAmi.valueAsString,
 				}),
 				instanceType: InstanceType.of(InstanceClass.G4DN, InstanceSize.XLARGE),
+				requireImdsv2: true,
```
philmcmahon, line 457: needs to be included in the other ASG as well

srbd (approved):

PR #153, opened 2025-07-22
Add warning text explaining that url transcription is out of service

philmcmahon:
## What does this change?
The media download service currently isn't working. It may take some time to work out a solution, so this PR disables url transcription in the UI with a big red error message to reduce user frustration

It looks like this:

Screenshot 2025-07-22 at 09 43 10

## How to test
Tested locally, I think this needs to go out sooner rather than later


srbd (approved):
good call

PR #155, opened 2025-07-23
Add volumes for directories needed by media download service

philmcmahon:
## What does this change?
I noticed recently when debugging the media download service that there are various warnings about it wanting to write to directories that are read only due to https://github.com/guardian/transcription-service/pull/114

This PR adds new volumes to remove those warnings.

## How to test
I've tested this on CODE and verified that the warnings stop


srbd (approved):
nice one

twrichards (approved):
good spot, changes make total sense. I assume you've tested in CODE that they have the desired effect.

PR #122, opened 2025-01-20
zip and unzip in progress

marjisound:
## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #167, opened 2025-08-22
Combine all transcription output text into a single JSON file

philmcmahon:
## What does this change?
This PR aims to make it easier to add new output formats to the transcription service, as well as generally simplifying how files are handled in the service.

Right now, when a transcript is requested the API generates 3 presigned urls for the srt, text and json output formats we request from whisper. These are passed via SQS and used by the worker to upload the output files. The output handler then downloads the files from these locations.

Handling 3 different urls for every transcript adds a bit of complexity.

In this PR I'm adding the concept of 'combinedOutput' which is a single json file containing the transcription output in every format. This would mean that the API (or Giant) just need to provide a single presigned url to the transcription service, the worker only needs to upload one file and the export service only needs to download one file. Happy times!

I've not removed the old functionality yet because:
 - there may be transcription jobs in progress at the point of deployment
 - we still need the txt/srt urls to support the 'download raw file' functionality on the export page. I'll reimplement this in a future PR to use the combined output.

This replaces the work started by @marjisound in https://github.com/guardian/transcription-service/pull/122  - sorry Marjan! I thought this would be less effort than dealing with a zip file. 

Following a discussion with @zekehuntergreen we decided to gzip the combined file. This adds some complexity - not so much in the zipping/unzipping but I decided to set the Content-Encoding header on the file in S3 (to gzip for the gzipped file).

The advantage of setting the content-encoding header in s3 is mainly to tell a potential browser that downloads the file how to open it. The disadvantage is that you have to set the header at the point of generating the presigned url, and the exact same header must be set for the upload. You can then only use that url to upload a file with the right content-encoding header set. 

TODO (in future prs):
 - Get the SRT/TXT downloads working with the new combined file
 - Remove the old 3 file logic

## How to test
I've tested this on CODE. There's a lot of zod

### DIFF
--- a/packages/api/src/export.ts
+++ b/packages/api/src/export.ts
@@ -44,18 +44,13 @@ export const getCombinedOutput = async (
 		config.app.transcriptionOutputBucket,
 		key,
 	);
+
 	if (isS3Failure(combinedOutputText)) {
-		if (combinedOutputText.failureReason === 'NoSuchKey') {
-			const msg = `Failed to export transcript - file has expired. Please re-upload the file and try again.`;
-			return {
-				status: 'failure',
-				failureReason: msg,
-			};
-		}
-		const msg = `Failed to fetch transcript. Please contact the digital investigations team for support`;
-		logger.error(
-			`Fetching from s3 failed, failure reason: ${combinedOutputText.failureReason}`,
-		);
+		const msg =
+			combinedOutputText.failureReason === 'NoSuchKey'
+				? `Failed to export transcript - file has expired. Please re-upload the file and try again.`
+				: `Failed to fetch transcript. Please contact the digital investigations team for support`;
+		logger.error(msg);
 		return {
 			status: 'failure',
 			failureReason: msg,
--- a/packages/api/src/export.ts
+++ b/packages/api/src/export.ts
@@ -58,7 +58,9 @@ export const getCombinedOutput = async (
 		};
 	}
 
-	const result = TranscriptionResult.safeParse(combinedOutputText);
+	const result = TranscriptionResult.safeParse(
+		JSON.parse(combinedOutputText.text),
+	);
 	if (result.success) {
 		return {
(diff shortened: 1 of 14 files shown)


packages/api/src/export.ts
```diff
@@ -25,40 +84,50 @@ export const exportTranscriptToDoc = async (
...
 	const transcriptS3Key = item.transcriptKeys[format];
-	const transcriptText = await getObjectText(
-		s3Client,
-		config.app.transcriptionOutputBucket,
-		transcriptS3Key,
-	);
-	if (isS3Failure(transcriptText)) {
-		if (transcriptText.failureReason === 'NoSuchKey') {
-			const msg = `Failed to export transcript - file has expired. Please re-upload the file and try again.`;
+	// use a mutable variable here as a temporary measure until the old non-combined format is removed
```
twrichards, line 91:
```suggestion
	// FIXME use a mutable variable here as a temporary measure until the old non-combined format is removed
```

packages/backend-common/src/sqs.ts
```diff
@@ -80,6 +80,19 @@ export const generateOutputSignedUrlAndSendMessage = async (
 		7,
 	);
 
+	const combinedOutputKey = `combined/${s3Key}.json`;
+
+	const combinedUrl = await getSignedUploadUrl(
```
twrichards, line 85: would be good if `getSignedUploadUrl` could be refactored to take a single parameter which is an object, so that the **many** inputs are forced to be labelled at the call site(s)

twrichards (approved):
code LGTM, few non-blocking comments - haven't tested myself, but I see you have in CODE 🙏

packages/backend-common/src/sqs.ts
philmcmahon: Agreed, it's quite confusing already. I'll do it in a follow up PR

PR #169, opened 2025-08-27
Always log out errors before throwing them

philmcmahon:
## What does this change?
It's often quite hard to trace problems in the transcription tool logs. This PR makes use of the structured logging setup to include the full error object in our logging, so that you don't have to try tie a stack trace to a particular error message in kibana. 

The transcription worker logs get a bit mangled (multi line logs get split into different log entries in elastic) on their way to ELK as they go to a file first and then to ELK using devx logs. I think we could probably sort it to log in a JSON format, but this is a convenient stop gap


srbd (approved):
error handling

PR #170, opened 2025-08-27
Tag GHCR whisper container with branch name

philmcmahon:
## What does this change?
Currently when baking the whisper AMI we use for the transcription service we download the container with tag `transcription-service:latest`. This means that we end up baking the most recent container that has been published - even if it was baked on a feature branch. 

Recently this caused a problem where a broken container was published, baked into an AMI, then we deployed `main` and the new AMI was picked up with the container baked from the broken feature branch.

To get round this, we should instead only use containers baked from the `main` branch in production. If we need to test a container on a feature branch then we will have to do that locally, or to create a temporary test recipe in amigo. 

This PR is the first step -tagging containers published to the github container registry with the branch they were baked on.

## How to test
I tested this on this feature branch! Looks like this:

Screenshot 2025-08-27 at 11 30 04

### DIFF
--- a/.github/workflows/build-whisper-container.yml
+++ b/.github/workflows/build-whisper-container.yml
@@ -57,6 +57,6 @@ jobs:
           # to add x86: linux/amd64
           platforms: linux/arm64
           push: true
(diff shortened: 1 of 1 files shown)


srbd (approved):
Just confirmed with Phil that the next step will be to change in the [bake]( https://amigo.gutools.co.uk/recipes/investigations-transcription-service) to `docker-pull` scoped to `main:latest` rather than just `latest`

PR #173, opened 2025-08-29
Pull localstack queue creation out into a separate script

philmcmahon:
## What does this change?

Even though we have localstack persistence enabled as per the docs here https://docs.localstack.cloud/aws/capabilities/state-management/persistence/ I find that I often have to recreate the local queues. It's annoying having to re run the entire setup script to achieve this, so this PR pulls the queue creation bit out into a separate script.

The dynamo table does seem to persist, so I've left that in setup.sh

## How to test
Works on my machine...


twrichards (approved):
LGTM

Is there any documentation on why we need FIFO queues?

philmcmahon:
@twrichards I remember having a discussion about it, but I can't think why we need them to be FIFO, since we always process the whole queue more or less immediately so ordering isn't really an issue. Maybe @zekehuntergreen might remember

PR #174, opened 2025-09-01
Use intel for whisper.cpp jobs. Upgrade whisper.cpp

philmcmahon:
## What does this change?
Recently, the bakes AMIgo was generating with whisper.cpp seemed to stop working - whisper would fail with `illegal instruction (core dumped)`. I haven't tried to debug this, instead I decided to create a new whisper.cpp container with the latest version of whisper.cpp. 

Unfortunately, I couldn't get this new version of whisper.cpp to run on a graviton instance. As an experiment I tried switching to intel, which worked fine - so I think we should just switch to intel.

Downsides:
 - higher costs
 - possibly worse transcription performance

However, we only use whisper.cpp to transcribe files shorter than 10 minutes (longer files run with whisperx on gpu instances), so these downsides won't have much of an impact. The alternative is to spend time trying to get whisper.cpp working on graviton, I think our time could be better spent experimenting with parakeet (https://medium.com/data-science-in-your-pocket/nvidia-parakeet-v2-the-smallest-fastest-free-speech-recognition-asr-model-28ee2ccfac51) or on the giant/transcription service integration.

I did try using the containers published here https://github.com/ggml-org/whisper.cpp/pkgs/container/whisper.cpp - it does say they publish one for linux/arm64, but I wasn't able to pull a container to a graviton instance

## How to test
I've tested this on CODE and verified that it can succesfully transcribe files.

### DIFF
--- a/containers/whisper.Dockerfile
+++ b/containers/whisper.Dockerfile
@@ -5,7 +5,9 @@ LABEL com.theguardian.transcription-service.whisper-container="Whisper.cpp conta
 RUN apt-get update
 RUN apt-get install -y ffmpeg wget git build-essential cmake
 RUN git clone https://github.com/ggerganov/whisper.cpp
-RUN cd whisper.cpp && git reset --hard v1.7.6 && cmake -B build && cmake --build build -j --config Release
+RUN cd whisper.cpp && git reset --hard v1.7.6
+RUN cd whisper.cpp && cmake -B build  -DWHISPER_NO_AVX=ON -DWHISPER_NO_AVX2=ON -DWHISPER_NO_FMA=ON -DWHISPER_NO_F16C=ON
+RUN cd whisper.cpp cmake --build build -j --config Release
 RUN bash /opt/whisper.cpp/models/download-ggml-model.sh tiny
 RUN bash /opt/whisper.cpp/models/download-ggml-model.sh medium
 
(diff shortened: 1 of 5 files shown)


packages/worker/src/transcribe.ts
```diff
@@ -361,7 +361,7 @@ export const runWhisper = async (
 		const result = await runSpawnCommand('transcribe', 'docker', [
 			'exec',
 			containerId,
-			'whisper.cpp/main',
+			'whisper.cpp/build/bin/whisper-cli',
```
philmcmahon, line 364: this is updated following the docs here https://github.com/ggerganov/whisper.cpp

containers/whisper.Dockerfile
```diff
@@ -3,9 +3,9 @@ WORKDIR /opt
 LABEL com.theguardian.transcription-service.whisper-container="Whisper.cpp container with models downloaded, including ffmpeg"
 
 RUN apt-get update
-RUN apt-get install -y ffmpeg wget git build-essential
+RUN apt-get install -y ffmpeg wget git build-essential cmake
 RUN git clone https://github.com/ggerganov/whisper.cpp
-RUN cd whisper.cpp && git reset --hard v1.5.5 && make
+RUN cd whisper.cpp && git reset --hard v1.7.6 && cmake -B build && cmake --build build -j --config Release
```
philmcmahon, line 8: updated base off https://github.com/ggerganov/whisper.cpp

PR #176, opened 2025-09-08
[Save to Giant part 1] Support running external jobs in media download service 

philmcmahon:
## What does this change?
This PR modifies the media download service so that it can process jobs coming from other services (in this case giant). The main change is that it needs to make use of a presigned url provided by the third party service to upload the output media to S3. 

Also:
 - new localstack queues for testing external media download jobs locally
 - I've pulled the CDK for the media download service out into a separate file as the existing cdk file was getting a bit big

This is needed for this feature in giant: https://github.com/guardian/giant/pull/283

## How to test
I've done one end to end test on playground, as well as verifying that the traditional media download service still works

### DIFF
--- a/packages/backend-common/src/s3.ts
+++ b/packages/backend-common/src/s3.ts
@@ -180,7 +180,7 @@ export const getObjectWithPresignedUrl = async (
 export const uploadObjectWithPresignedUrl = async (
 	presignedUrl: string,
 	metadata: MediaMetadata,
-) => {
+): Promise<void> => {
 	logger.info(
 		`Uploading with ${presignedUrl}, metadata ${JSON.stringify(metadata)}`,
 	);
@@ -203,7 +203,6 @@ export const uploadObjectWithPresignedUrl = async (
 		if (response.ok) {
 			const text = await response.text();
 			logger.info(`File uploaded successfully, response: ${text}`);
-			return true;
 		} else {
 			throw new Error(
 				`Upload failed with status ${response.status}, message ${response.statusText}`,


hoyla:
Child of https://github.com/guardian/giant/issues/311

packages/backend-common/src/s3.ts
```diff
@@ -174,6 +177,44 @@ export const getObjectWithPresignedUrl = async (
...
+			headers: {
+				'Content-Type': 'application/octet-stream',
+				'Content-Length': contentLength.toString(),
+			},
+		} as RequestInit);
+
+		if (response.ok) {
+			const text = await response.text();
+			logger.info(`File uploaded successfully, response: ${text}`);
+			return true;
```
twrichards, line 206: [minor] this function never returns false (only returns true or it throws) so perhaps this return is redundant

packages/cdk/lib/transcription-service.ts
twrichards, line 1: adding cdk snapshot tests (in a pre-requisite PR) would be useful here, to see the resulting diff in actual infra changes

twrichards: also what's the benefit of moving to its own file (abstracting it via `makeMediaDownloadService`) are you planning on producing multiple media-download-services?

scripts/create-local-queues.sh
```diff
@@ -44,7 +44,7 @@ OUTPUT_QUEUE_URL_LOCALHOST=${OUTPUT_QUEUE_URL/sqs.eu-west-1.localhost.localstack
 
 echo "Created output queue in localstack, url: ${OUTPUT_QUEUE_URL_LOCALHOST}"
 
-MEDIA_DOWNLOAD_QUEUE_URL=$(aws --endpoint-url=http://localhost:4566 sqs create-queue --queue-name=$APP_NAME-media-download-queue-DEV | jq .QueueUrl)
+MEDIA_DOWNLOAD_QUEUE_URL=$(aws --endpoint-url=http://localhost:4566 sqs create-queue --queue-name=$APP_NAME-media-download-queue-DEV.fifo --attributes "FifoQueue=true,ContentBasedDeduplication=true" | jq .QueueUrl)
```
twrichards, line 47: why does the queue need to be FIFO?

twrichards (approved):
LGTM - a few comments/questions

scripts/create-local-queues.sh
philmcmahon: In this case I'm just bringing it in line with the existing infra (which was already fifo - so the local infra should have been https://github.com/guardian/transcription-service/pull/176/files#diff-4c04d9d03ec555fc8ffc3cdb2814f0e13bb54815a58ab0160ed1182e272ed9a5) but I think that it shouldn't be FIFO really, something to change in a follow up bit of work

PR #178, opened 2025-09-19
Add alarms for multiple dead letter queue visible messaages thresholds

philmcmahon:
## What does this change?
The existing dead letter queue alarm we have has been really useful for getting a warning when a trasncription job has failed. However, if multiple jobs fail and keep failing so that there is always > 1 message in the dead letter queue, then the alarm only fires once. 

To resolve this I've added 3 new alarms that fire at 5, 10 and 20 messages. I've also made the alarm names more readable so we should be able to distinguish betewen these different levels.

My original plan was to use the 'SentMessages' but that is [not accurate for dead letter queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-available-cloudwatch-metrics.html#dlq-cloudwatch-metrics), so I think this is the only viable solution without creating a lambda or similar to do the monitoring 
## How to test
Much like https://github.com/guardian/transcription-service/pull/177, this needs to be tested on PROD as we only create alarms on PROD

### DIFF
--- a/packages/cdk/lib/alarms.ts
+++ b/packages/cdk/lib/alarms.ts
@@ -23,26 +23,34 @@ export const makeAlarms = (
 	const oldestMessageAlarmThresholdMinutes = 60;
 	const oldestMessageAlarmThresholdSeconds =
 		oldestMessageAlarmThresholdMinutes * 60;
-	const alarms = [
+
+	// This isn't very nice but since the NumberOfMessagesSent is not accurate for dead letter queues, we create multiple
+	// alarms at different thresholds to ensure we get multiple alarms when there is an ongoing incident
+	// (see https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-available-cloudwatch-metrics.html)
+	const dlQueueAlarms = [1, 5, 10, 20].map((threshold) => {
 		// alarm when a message is added to the dead letter queue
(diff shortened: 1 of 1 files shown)


packages/cdk/lib/alarms.ts
```diff
@@ -23,26 +23,34 @@ export const makeAlarms = (
...
+	// alarms at different thresholds to ensure we get multiple alarms when there is an ongoing incident
+	// (see https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-available-cloudwatch-metrics.html)
+	const dlQueueAlarms = [1, 5, 10, 20].map((threshold) => {
 		// alarm when a message is added to the dead letter queue
 		// note that queue metrics go to 'sleep' if it is empty for more than 6 hours, so it may take up to 16 minutes
 		// for this alarm to trigger - see https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-monitoring-using-cloudwatch.html
-		new Alarm(scope, 'DeadLetterQueueAlarm', {
-			alarmName: `transcription-service-dead-letter-queue-${scope.stage}`,
+		return new Alarm(scope, `DeadLetterQueueAlarmThreshold${threshold}`, {
+			alarmName: `Transcription service ${scope.stage}: ${threshold} messages in DLQ`,
```
zekehuntergreen, line 35: Good thinking making the names more readable

zekehuntergreen (approved):

PR #201, opened 2025-10-14
Add webpage snapshot service

philmcmahon:
## What does this change?
This adds a new service to the transcription-service - 'webpage-snapshot'. This actually has nothing to do with transcription, but we decided the transcription service was a good home for it, given that's where the other giant 3rd party services live (transcription, media download). 

webpage-snapshot is a lambda function that runs [puppeteer](https://pptr.dev/) in order to generate a screenshot of a page, as well as a copy of the html for the page. These are returned to giant in a single combined JSON file, with the screenshot encoded in base64.

The main complexity with the service is that it relies on chromium, and the default bundled version of chromium that comes with puppeteer doesn't work on aws lambda. Instead we use a lambda layer provided by https://github.com/Sparticuz/chromium. Further complexity is added as, locally, we need to run the bundled version of chromium - note the dynamic imports in index.ts

The webpage snapshot service has its own task queue, currently wired up to an sns topic so that giant can trigger multiple 'archive' services with a single write to SNS. The service reuses the existing 'giant media download output' queue - ultimately this should be renamed to 'giant remote ingest output' to reduce confusion, but I wanted to reduce the number of moving parts in this PR. 

## How to test
This is currently live on playground.

### DIFF
--- a/packages/media-download/src/index.ts
+++ b/packages/media-download/src/index.ts
@@ -208,8 +208,12 @@ const main = async () => {
 			const tJob = TranscriptionMediaDownloadJob.parse(parsedInput);
 			await reportDownloadFailure(config, sqsClient, tJob);
 		} else if (isExternalMediaDownloadJob(job)) {
-			const eJob = ExternalUrlJob.parse(parsedInput);
-			await reportExternalFailure(eJob, sqsClient, ytDlpResult.errorType);
+			const externalJob = ExternalUrlJob.parse(parsedInput);
+			await reportExternalFailure(
+				externalJob,
+				sqsClient,
+				ytDlpResult.errorType,
+			);
 		}
 	} else {
 		if (isTranscriptionMediaDownloadJob(job)) {
@@ -228,12 +232,12 @@ const main = async () => {
 				ytDlpResult.metadata,
 			);
 		} else if (isExternalMediaDownloadJob(job)) {
-			const eJob = ExternalUrlJob.parse(parsedInput);
+			const externalJob = ExternalUrlJob.parse(parsedInput);
 			await uploadObjectWithPresignedUrl(
-				eJob.mediaDownloadOutputSignedUrl,
+				externalJob.mediaDownloadOutputSignedUrl,
 				ytDlpResult.metadata.mediaPath,
 			);
-			await reportExternalJob(eJob, sqsClient, ytDlpResult.metadata);
+			await reportExternalJob(externalJob, sqsClient, ytDlpResult.metadata);
 		}
 	}
 };


packages/backend-common/src/s3.ts
```diff
@@ -179,16 +178,14 @@ export const getObjectWithPresignedUrl = async (
...
 export const uploadObjectWithPresignedUrl = async (
 	presignedUrl: string,
-	metadata: MediaMetadata,
+	path: string,
 ): Promise<void> => {
-	logger.info(
-		`Uploading with ${presignedUrl}, metadata ${JSON.stringify(metadata)}`,
-	);
-	const fileStream = createReadStream(metadata.mediaPath);
+	logger.info(`Uploading with ${presignedUrl}, file ${path}`);
```
twrichards: whilst these are our logs, the pre-signed url contains a secret (and is also very long), so wonder if its worth redacting it a bit? perhaps putting the path first in the message too, if the filename is valuable?

packages/cdk/lib/transcription-service.ts
```diff
@@ -714,6 +737,85 @@ export class TranscriptionService extends GuStack {
...
+
+		webpageSnapshotLambda.addEventSource(
+			new SqsEventSource(webpageSnapshotQueue),
+		);
+
+		webpageSnapshotLambda.addToRolePolicy(getParametersPolicy);
+		webpageSnapshotLambda.addToRolePolicy(putMetricDataPolicy);
+
+		webpageSnapshotLambda.addToRolePolicy(giantRemoteIngestOutputSendPolicy);
+
```
twrichards: minor: might be possible to add these in the config of the lambda itself, via `initialPolicy` array

twrichards (approved):
awesome work @philmcmahon - I've left some comments, though none of them blocking I think, so approving 😄

packages/backend-common/src/s3.ts
philmcmahon: ah agreed it shouldn't be logged. It was helpful logging it when trying to debug why the uploads were failing - but that can all be done locally.

packages/cdk/lib/transcription-service.ts
```diff
@@ -112,6 +114,22 @@ export class TranscriptionService extends GuStack {
 			},
 		).valueAsString;
 
+		const giantRemoteIngestOutputQueueArn = new GuStringParameter(
+			this,
+			'mediaDownloadGiantOutputQueueArn', // TODO: Should be called remote ingest giant output
+			{
+				fromSSM: true,
+				// TODO: Should be called remote ingest giant output
```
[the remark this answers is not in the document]
philmcmahon: Good point, I've renamed it, created the new parameter and added a 'DELETE ME' with a link to this PR to the description of the old parameter

packages/cdk/riff-raff.yaml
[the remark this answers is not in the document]
philmcmahon: It's not generated by gucdk, because we (were) using spot instances via a mixed instances policy, we had to use a generic autoscaling group rather than a guautoscalinggroup, which meant that gucdk couldn't generate the riffraff.yaml for the worker asg (dates back to this PR https://github.com/guardian/transcription-service/pull/4). 

We're currently not making use of the mixed instances policy so if we were to abandon spot instances entirely then we could go back to using a guautoscalinggroup and autogenerate this file I think

PR #218, opened 2025-11-12
always have a minimum of one GPU worker

twrichards:
Now we have #215, #216 & #217 released we're in a good place to know/understand the benefit (to user experience of transcription service) of having a minimum of one GPU worker - primarily to avoid warm-up time (both the box visible `SecondsFromEnqueueToStart` metric and the WhisperX warm-up time visible via `SecondsForWhisperXStartup` metric).

So this PR does that (both in the CDK infra definition and in the `worker-capacity-manager`).


philmcmahon:
I think we still want to scale to 0 on CODE

PR #228, opened 2025-12-03
Generate a user agent when running yt-dlp

philmcmahon:
## What does this change?
As suggested by @twrichards , use  https://www.npmjs.com/package/@imaginerlabs/user-agent-generator to make the media download service look a bit more human. 

## How to test
Tested on CODE. Only time will tell the impact of this


dominickendrick (approved):
🏃 🤖

PR #235, opened 2025-12-05
Fix translation for combined transcribe/translate jobs from giant

philmcmahon:
## What does this change?
I think, ever since we switched to using whisperx for transcription, translation has been broken in giant. This is because language detection was broken for the 'combine transcribe and translate' job when using whisperx. Giant currently doesn't provide language codes to the transcription service (because we don't request it from the user) so relies on language detection

What was broken:
 - the language code gets output to stdout, we were trying to read it from stderr. 
 - Even if the language code was detected, we were overriding it with the result of calling getLanguageCode, which just returned 'auto' for whisperx* 
 
This meant:
 - translation wasn't working at all for giant, as it doesn't provide a language code so relies on language detection
 - every job sent from giant to whisperx was getting 'translated' even if it was in english
 
To resolve these issues I have:
 - deleted the getLanguageCode function - it might be useful if we bring back whisper.cpp but I think that is unlikely, and right now it just creates unnecessary confusion. In any case, whisper.cpp isn't so slow that running language detection as a separate task is a huge win.
 - Fixed extractWhisperXStderrData - it is now extractWhisperXStdoutData
 - Added some logic so that we always respect a user provided languageCode in the combineTranscribeAndTranslate job, but otherwise fall back to the detected language code of the first transcription job

*_as language detection isn't worth doing in whisperx as it would take a similar amount of time as transcribing the whole file._

## How to test
Tested on playground and verified:
 - english video only gets transcribed once
 - spanish song gets properly translated

### DIFF
--- a/packages/worker/src/transcribe.ts
+++ b/packages/worker/src/transcribe.ts
@@ -186,13 +186,20 @@ const transcribeAndTranslate = async (
 			metrics,
 		);
 
+		// translation only works in whisperx if we know the language code. Also, we don't want to translate files that
+		// are already in english. If there is a user provided language, trust that. Otherwise, rely on the detected
+		// language from the transcript job
+		const languageCodeOrDetected =
+			languageCode !== 'auto'
+				? languageCode
+				: transcription.metadata.detectedLanguageCode;
+
 		const translation =
-			languageCode === 'en' ||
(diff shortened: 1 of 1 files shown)


twrichards (approved):
🕵️  great work @philmcmahon - not tested myself but I see you have 👏

PR #238, opened 2025-12-11
WIP/EXPERIMENT run unlimited transcriptions in parallel (also use long-polling)

twrichards:
Co-authored-by: @philmcmahon 

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #241, opened 2025-12-15
Perform translation in the same job as transcription

### LINKED ISSUE #240: Run transcription and translation in a single worker job
This already applies for giant. The behaviour would be:

unknown language: transcribe, get language, translate
known non-english language: transcribe and translate in parallel

philmcmahon:
## What does this change?
Currently, when a user requests a translation of their transcription task, the transcription API just sends two jobs to the queue, one for the transcription, the other for the translation. The user then receives 2 emails - one for each job.

This PR modifies that behaviour so that transcripion and translation are performed in a single job on the same worker. This was already the behaviour for jobs coming from giant, and issues such as this one https://github.com/guardian/transcription-service/pull/235 have made me think that it would be preferable to have fewer pathways through the app rather than separate flows for giant/transcription service. 

The main interesting change in this pr is in the `transcribeAndTranslate` function - which now will attempt to run the translation job in parallel with the transcription job (based off learnings from https://github.com/guardian/transcription-service/pull/238). Otherwise, it's a lot of rewiring in the output handler, viewer and export page to account for the user now only getting one email when the job(s) finish.  

Note: Based on [some testing](https://docs.google.com/spreadsheets/d/1P1Vq9ltKNHFcTMs6Ak9C04RLEVWOfjxV57wYcZf7ODg/edit?gid=0#gid=0), we think that running 2 jobs in parallel on the instance type we're using has a negligible affect on run time - the job takes about twice as long but we get two jobs done at once. Doing more than 2 jobs at a time with our current batch size risks running out of GPU memory.

## How to test
I've tested this locally and on CODE. Attempting to transcribe any foreign language file and asking for a translation would be a suitable test

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4313,14 +4313,6 @@
 			"dev": true,
 			"license": "BSD-3-Clause"
 		},
-		"node_modules/@imaginerlabs/user-agent-generator": {
-			"version": "1.0.2",
-			"resolved": "https://registry.npmjs.org/@imaginerlabs/user-agent-generator/-/user-agent-generator-1.0.2.tgz",
-			"integrity": "sha512-QTOi46xtGLnYG4OFxyaoRL+KnL6RopZ5Foh2dO+OVnoYpv5wM9zhSlgiOwr43I0Q6cIeiDQ/8xQDmXY7f4As1A==",
-			"engines": {
-				"node": ">=14.0.0"
-			}
-		},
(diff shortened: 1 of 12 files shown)


twrichards (approved):
great work - think we might benefit from porting the long-polling logic here (as per https://github.com/guardian/transcription-service/pull/238#issuecomment-3723017023)

hoyla:
Fixes #240

PR #254, opened 2026-01-23
Fix translation logic in worker (again) + write some unit tests 😭 

philmcmahon:
## What does this change?
There's currently a bug in the transcription tool where, if no language code is specified (it is set to 'auto') - which is true of all files coming from giant - the file will gets transcribed _and_ translated, even if during the transcription we detect that the source language is english.

I've fixed this with a new 'getTranslationConfig' function which handles all the logic about deciding whether to translate a file or not and what language code to use. As this is the third time I've had to mess with this logic, I've added some unit tests to document the expected behaviour

Some history on this:
 - https://github.com/guardian/transcription-service/pull/235
 - https://github.com/guardian/transcription-service/pull/241
 - github.com/guardian/transcription-service/pull/234

## How to test
Run a transcription job from giant for a file in english, you should only see whisperx get run once

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -85,8 +85,9 @@ jobs:
       - name: Package API/Client/Update
         run: |
           npm run package
-      - name: Package Worker
+      - name: Run tests and package Worker
         run: |
+          npm run worker::test
           npm run worker::build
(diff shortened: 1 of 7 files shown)


twrichards (approved):
lovely stuff

PR #276, opened 2026-04-09
Pin and tidy dependencies, remove axios, bump node

philmcmahon:
## What does this change?
Following recent recommendations by DevX following the axios exploit, this PR:
 - Pins all our dependencies to a specific version
 - Removes unnecessary dependencies where they are duplicated (e.g. @types/node which is in backend-common so doesn't need to be in any of the other backend apps)
 - Removes axios - we were only using it in one place
 - Bumps node to 24 (latest LTS)
 - Applies cooldown dependabot config
 - Applies min-release-age flag added in https://github.com/npm/cli/pull/8965/changes

Tested on CODE

## Why?
image


twrichards (approved):
big diff (so not reviewed every line) but change looks very sensible from description and you've tested on CODE 🙏 

great change

PR #277, opened 2026-04-09
Remove whisper cpp

philmcmahon:
## What does this change?
This removes whisper.cpp from the transcription service. Whisper.cpp is really good, with a million times faster startup time than whisperx, but we are reliant on the diarization capabilities of whisperx and having two pipelines adds unnecessary complexity 

Tested on CODE


twrichards (approved):
love to see code get deleted

PR #280, opened 2026-04-15
Fetch models from S3 rather than baking into the AMI

philmcmahon:
## What does this change?
Whilst working on adding llama.cpp to our transcription workers I learned that it is not recommended to have large model files baked into the AMI:
 - https://medium.com/analytics-vidhya/do-not-put-your-ml-models-in-aws-ebs-3ae0db559669
 - https://docs.aws.amazon.com/ebs/latest/userguide/initalize-volume.html

Basically, EBS volumes are really just S3 so the instance can start up but with more sync work to do in the background. It sounds like you can speed it up a bit by [setting a provisioned rate for volume initialization](https://docs.aws.amazon.com/ebs/latest/userguide/initalize-volume.html#volume-initialization-rate). 

Even if it wasn't for the above, we are also not making use of the 125GB NVMe volume that comes for free with a g4dn.xlarge instance. This should be significantly faster than using EBS.

This PR:
 - adds a github action to fetch whisperx models (and Qwen 3B) and to push them to S3
 - Modifies the transcription worker startup to fetch the models from S3 rather than rely on them being baked into the AMI

In my testing this delivered a roughly 5 minute faster startup time for whisperx when booting an instance from scratch. Hopefully there will also be some performance benefits for PROD where we have an instance always on as the model will be being read from the NVMe drive rather than EBS.

Some notes:
 - As well as whisperx models, this also fetches Qwen3B in preparation for use of that model with llama.cpp in the future
 - the nvme drive gets mounted automatically thanks to our use of the deep learning AMI
 - As a follow up we need to remove the download models step from the AMI bake https://github.com/guardian/amigo/blob/main/roles/whisperx/tasks/main.yml#L27
 - I discovered that the riffraff artifact action doesn't support artifacts > 2GB. This resolves that, and needs merging first https://github.com/guardian/actions-riff-raff/pull/306 - in the meantime I'm currently uploading to S3 directly from the github action

Depends on https://github.com/guardian/transcription-service/pull/277 and https://github.com/guardian/investigations-platform/pull/669 (as well as https://github.com/guardian/actions-riff-raff/pull/306 mentioned above)

Tested on CODE


twrichards (approved):
awesome work 🙌

PR #281, opened 2026-04-15
Add github action to build llama.cpp

philmcmahon:
## What does this change?
We've got an interest in adding more LLM-based capability to the transcription service, in particular for translation.

https://github.com/ggml-org/llama.cpp seems like a reasonably good tool for this - it includes `llama-server` which provides "[OpenAI API](https://github.com/openai/openai-openapi) compatible chat completions, responses, and embeddings routes"

Unfortunately for reasons I don't quite understand they don't provide prebuilt binaries to work with Nvidia CUDA GPUs, so we have to build it ourselves. This adds a github action to build llama.cpp along with a riffraff aws-s3 deploy step to upload it to the amigo-data bucket from where it can be baked into an AMI. (Given the findings in https://github.com/guardian/transcription-service/pull/280 it might make sense to just fetch llama.cpp when the app starts up, but it is only 270MB so I don't think the EBS sync slowness won't be a huge delaying factor - in my testing it started up within a few seconds.

The compile process is slooow. Originally I did this as an AMIgo role but it was taking forever and regularly timing out so I gave up on that, I think it's nicer to have it living in github actions anyway as the developer experience for AMIgo roles that you can't test locally with multipass is terrible.

Note that llama-server will only work if we set LD_LIBRARY_PATH to `install/lib` - we might be able to use static linking instead to avoid this extra setting but I think it's fine for now.

I've tested this by baking it into this amigo with this role https://github.com/guardian/amigo/pull/1809 and then booting a transcription service instance using the resulting AMI and checking that llama-server works.


twrichards (approved):
awesome

PR #283, opened 2026-04-20
Set volume initialisation rate on worker instances

philmcmahon:
## What does this change?
This PR sets a volume initialization rate of 300MB/s (the maximum) for our worker instances. See here for details of what this is: https://docs.aws.amazon.com/ebs/latest/userguide/initalize-volume.html#volume-initialization-rate

The reason it helps us is that the transcription service worker AMI is large due to the CUDA drivers and (at the moment) the baked-in model files. Currently it is taking around 15 minutes for an instance to 'warm up' whilst EBS syncs all this data from S3, this cuts the time to around 4 minutes.

This is quite expensive - $0.0036 per GB. With that in mind, this change drops the (unused) small/large whisper models from the AMI (we can drop these completely after https://github.com/guardian/transcription-service/pull/280) 

To further cut the cost I've done this https://github.com/guardian/amigo/pull/1828 which should shave around 30GB off our AMI size.

That still leaves 40GB, so I think enabling this will cost $0.14 per instance start. Where the instance is an 'extra' box (not our reserved instance) then this will save us $0.04 due to the job finishing 10 minutes earlier (GPU instances are $0.52/hour). 

Over the last 30 days we have launched 300 boxes, at that rate this change would cost us $40ish/month.

I've tested this on CODE and verified the faster start times

### DIFF
--- a/packages/cdk/lib/__snapshots__/transcription-service.test.ts.snap
+++ b/packages/cdk/lib/__snapshots__/transcription-service.test.ts.snap
@@ -1124,7 +1124,7 @@ exports[`The TranscriptionService stack matches the snapshot 1`] = `
             {
               "DeviceName": "/dev/sda1",
               "Ebs": {
-                "VolumeInitializationRate": 1000,
+                "VolumeInitializationRate": 300,
                 "VolumeSize": 100,
               },
             },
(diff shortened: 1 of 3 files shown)


twrichards (approved):
brill. does your $40/month estimate include the savings in running costs from the improved warm-up time (appreciate it's not as much as the increased cost of this change, but does offset somewhat)

philmcmahon:
> brill. does your $40/month estimate include the savings in running costs from the improved warm-up time (appreciate it's not as much as the increased cost of this change, but does offset somewhat)

A good point that's not very clear in the description. The improved warm up time should save us $0.04 per instance launch, so for 300 launches that would be $30 rather than $40, I imagine it will end up being $30-35ish as some of the launches will be deploys

PR #284, opened 2026-04-21
Add llama.cpp to transcription service so it can translate text and make arbitrary LLM requests

philmcmahon:
## What does this change?
This PR is part of a project that moves the transcription service away from basically being a wrapper around whisperx and into a more general "do things that need a GPU instance" service. In this PR I'm adding new functionality to the transcription worker where it can call out to either AWS Bedrock or a locally hosted llama.cpp server instance in response to prompts submitted by a user.

Key changes:
 - Worker changes so that it does LLM jobs as well as transcription jobs (this is the hardest to review as a lot of stuff moved around)
 - As part of the above - new llama-cpp.ts and llama-server.ts files to handle starting/stopping llama.cpp (note that we have to start/stop on every request rather than just having the server running constantly because it uses around 10GB of video ram, whisperx needs around 8 and we only have 16 in total)
 - New API endpoints: POST /llm-prompt for submitting a request and GET /llm-prompt for getting the result. These do all the S3/SQS faffery that is required to send the job to the worker
 - New /prompt frontend for trying out different prompts
 - New /translate frontend which works more or less the same as /prompt but it has a prebaked system prompt for translation
 - local harness for running llama.cpp locally (requires re-running setup.sh to install llama.cpp and save a model to /etc/gu)
 - As well as the llama.cpp stuff, this also supports sending prompts to bedrock - this will cost more money but allows us to use a much larger model.

This change depends on various bits of work to get llama.cpp installed on the worker instances:
 - https://github.com/guardian/transcription-service/pull/280
 - https://github.com/guardian/transcription-service/pull/281
 - https://github.com/guardian/amigo/pull/1809

Tested on CODE

The long term plan is to have a translation engine that lives entirely in our own infrastructure that we can wire giant up to - this bit just makes it easier to test stuff (and was quite fun to do)


packages/client/src/components/Translate.tsx
```diff
@@ -0,0 +1,249 @@
...
+	type LlmBackend,
+	LlmResult,
+} from '@guardian/transcription-service-common';
+import { QWEN3_LANGUAGES } from '@/components/languages';
+import { BackendPicker } from '@/components/PromptField';
+import { getResult, submitPrompt } from '@/services/llm';
+
+const POLL_INTERVAL_MS = 3000;
+
+const AUTO_DETECT = '';
```
philmcmahon: ?

twrichards: does this need addressing?

twrichards (approved):
awesome work @philmcmahon  🙌 

bit too much to review thoroughly but AFAICT it LGTM

PR #286, opened 2026-04-23
Stop API downloading the submitted file

philmcmahon:
## What does this change?
Back when we had 2 engines for transcription (whisper.cpp and whisperx) we implemented this https://github.com/guardian/transcription-service/pull/136 to get the duration of the submitted media file so we could decide which engine to use for the job. 

We have since removed whisper.cpp so this feature is redundant, and as a result we no longer need to be downloading the submitted media file before sending it for transcription. This PR removes that step

This should significantly improve the response time of the /transcribe-file endpoint - we have been seeing timeouts there

Needs testing on CODE

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -19,7 +19,6 @@ import {
 	getS3Client,
 	sendMessage,
 	writeDynamoItem,
-	downloadObject,
 	getYoutubeEventItem,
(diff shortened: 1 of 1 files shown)


twrichards (approved):
nice improvement (should be quicker & cheaper as result too)

PR #289, opened 2026-04-27
Allow bedrock calls from transcription API

philmcmahon:
## What does this change?
Following from https://github.com/guardian/transcription-service/pull/284 this PR makes it possible for the API to send prompt requests directly to bedrock rather than going via the transcription worker, resulting in a significant performance boost - everything can be done synchronously so that by the time the request returns we have a result.

Most of this PR is moving stuff around so that it's available in backend-common rather than just in the output-handler. I've also pulled some stuff out of the api index.ts file, which is getting a bit massive

Tested on CODE

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3144,6 +3144,126 @@
 				"node": ">= 10"
 			}
 		},
+		"node_modules/@next/swc-darwin-x64": {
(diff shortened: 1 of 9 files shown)


srbd (approved):
nicely synchronous

PR #293, opened 2026-06-10
Pin localstack version to version prior to license change

philmcmahon:
## What does this change?

I pulled the latest localstack today and found that it wouldn't work without a license key. Blog post here https://blog.localstack.cloud/the-road-ahead-for-localstack/ 

Our options are:
 - buy a license
 - switch to a different product for local queues and dynamo tables
 - just use real AWS resources for DEV mode

Or do this - pin to an old version of localstack and put off the problem for another day

Also: Remove deprecated 'version' property that results in an annoying message every time you run docker compose.

Tested on my machine, verified that localstack starts without a license key

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,9 +1,7 @@
-version: '3.8'
-
 services:
   localstack:
     container_name: '${LOCALSTACK_DOCKER_NAME:-localstack-main}'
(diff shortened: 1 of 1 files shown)


twrichards (approved):
makes sense - can we go to Garage?

PR #296, opened 2026-07-01
Support translation jobs from Giant

philmcmahon:
## What does this change?
This is the companion PR to https://github.com/guardian/giant/pull/780 - which adds a translation extractor to Giant. This builds on https://github.com/guardian/transcription-service/pull/284 - which added general LLM capability to the transcription service.

There are two parts to this PR - reorganising of the transcription service types that are shared with Giant and supporting the translation of arbitrary length text blobs.

### Shared types with Giant
If you take a look in `worker-interface.ts` you can see all the models the transcription service shares with giant. This PR:
 - bumps the version of Zod we're using (originally with the intention of using Zod to generate JSON schema files) - this required setting `"skipLibCheck": true` in the cdk project
 - Moves all types that are shared with giant into `worker-interface.ts`
 - Adds a lot of new types to handle the translation - notably TranslationTask
 - Add ts-node and nodemon to root project as they are needed in most projects and were being hoisted from packages/cdk

### Translation support
Key changes:
 - Update `llm.ts` to enforce 'no think' mode for bedrock and set a timeout so we don't wait endlessly for requests that have crashed at the bedrock end
 - set a `maxTokens` limit so we fail early on requests that are too large
 - Transcription service worker now reads all '[message attributes](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html)' from incoming messages and forwards an allow list (FORWARDED_MESSAGE_ATTRIBUTE_NAMES) set of names to output services. This allows giant to match jobs up with the relevant extractor without polluting the LlmTranslationJob model with giant specific stuff

### llama-server refactor
Key changes:
 - add various params in an attempt to make it go faster (see comments)
 - Refactor so that we don't stop the server after every job, instead only stop it when we have an incoming transcription job

### Translation engine
 - Add new `llm-translate` jobType which expects a TranslationTask
 - Two new libraries `langchain/textsplitters` and `js-tiktoken` to use to split translation text into suitably sized chunks, and to estimate how many tokens a given bit of text will consume
 - `worker/src/llm.ts` which is responsible for splitting the user prompt of an LlmPrompt up into chunks. Where splitting takes place, we update the system prompt to tell the LLM this is happening to discourage it from adding extra crap before/after the translation
 - `reduce-tokens.ts` which I added after discovering that translating PDF versions of wikipedia pages was extremely slow despite there not being that much text to translate. I think it's because of all the urls in the page. To try and alleviate this we replace all the urls with `(1)`, `(2)` etc using CJK brackets which are hopefully unlikely to appear in the text for translation otherwise, then add the urls back in post translation. **Note - maybe we should just strip urls entirely and not include them in the translation output, but this seems nicer for now**

## AI usage in this PR
The files `worker/src/llm.ts`  was heavily inspired by copilot - mainly the decision to use the libraries js-tiktoken and langchain/textsplitters. I did a lot of tidying up the generated code.

I also got copilot to sort out the 'forwarding message attributes' stuff because I didn't want to thread all that through myself.

I have tested:
 - Transcription service still works
 - I can run a prompt, run a transcription job and then run another prompt
 - Giant can still transcribe files
 - Files in giant get a translation added to them when uploaded


packages/backend-common/src/llm.ts
```diff
@@ -9,14 +9,20 @@ import {
...
 import { logger } from '@guardian/transcription-service-backend-common';
 import { LlmPrompt } from '@guardian/transcription-service-common';
 
 export const sendPromptToBedrock = async (
 	prompts: LlmPrompt,
 	bedrockModelId: string,
 ): Promise<string> => {
-	const client = new BedrockRuntimeClient({ region: 'eu-west-1' });
+	const client = new BedrockRuntimeClient({
+		region: 'eu-west-1',
```
twrichards: could this move to constants file for all aws config/defaults

packages/backend-common/src/sqs.ts
```diff
@@ -84,7 +106,7 @@ export const generateOutputSignedUrlAndSendMessage = async (
 		'gzip',
 	);
 
-	const engine = TranscriptionEngine.WHISPER_X;
+	const engine: TranscriptionEngine = 'whisperx';
```
twrichards, line 109: could this change back to a constant

packages/worker/package.json
```diff
@@ -15,12 +15,15 @@
 	"dependencies": {
 		"@aws-sdk/client-auto-scaling": "^3.1000.0",
 		"@aws-sdk/client-bedrock-runtime": "^3.1000.0",
-		"@guardian/transcription-service-backend-common": "1.0.0"
+		"@guardian/transcription-service-backend-common": "1.0.0",
+		"@langchain/textsplitters": "1.0.1",
+		"js-tiktoken": "1.0.21"
```
twrichards, line 20: is the worker connected to the internet? if so, concern would be that this lib could exfiltrate sensitive stuff

(IRL answer: no it shouldn't be)

twrichards (approved):
👏

PR #298, opened 2026-07-13
Scale down CODE workers at 8pm

philmcmahon:
## What does this change?
As a cost saving measure, this adds a job to scale down CODE workers at 8pm UTC, following a particularly expensive weekend leaving a GPU instance running.

I checked the build passes, I think we could just merge this and check for any errors in the morning.

### DIFF
--- a/packages/cdk/lib/transcription-service.ts
+++ b/packages/cdk/lib/transcription-service.ts
@@ -33,6 +33,7 @@ import {
 } from 'aws-cdk-lib';
 import { EndpointType } from 'aws-cdk-lib/aws-apigateway';
 import {
+	Schedule as AsgSchedule,
 	AutoScalingGroup,
 	BlockDeviceVolume,
 	GroupMetrics,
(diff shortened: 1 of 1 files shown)


packages/cdk/lib/transcription-service.ts
```diff
@@ -547,6 +548,17 @@ export class TranscriptionService extends GuStack {
...
 		);
 
+		// We frequently set the min workers on CODE to 1 so we can test stuff, then forget to scale it down
+		// Here we enforce an 8pm scale down, which might occasionally be annoying but is probably on balance
+		// less annoying than running up a $50 aws bill over the weekend because we forgot to scale down the workers
+		if (this.stage === 'CODE') {
+			transcriptionGpuWorkerASG.scaleOnSchedule('ScaleDownAtNight', {
+				schedule: AsgSchedule.cron({ hour: '20', minute: '0' }),
+				minCapacity: 0,
+				desiredCapacity: 0,
```
twrichards, line 558: I wouldn't be against adding `maxCapacity: 0,` just so when returning to it, there is no trace of how big it was before (keeping in mind how the UI just mindlessly ups the max whenever you type in a large desired capacity)
```suggestion
				minCapacity: 0,
				maxCapacity: 0,
				desiredCapacity: 0,
```

twrichards (approved):
very sensible - nice one @philmcmahon

packages/cdk/lib/transcription-service.ts
philmcmahon: sorry I missed this in my haste to merge. I think we want to keep max as something that is not zero so that jobs that get submitted will cause a worker to spin up (the capacity manager respects the maxCapacity), though if I hadn't blindly merged this I probably would get it to at least set the maxCapacity to whatever the default is for the asg

PR #301, opened 2026-07-14
Translation benchmarking/performance

philmcmahon:
## What does this change?
This PR builds directly on https://github.com/guardian/transcription-service/pull/296 with an aim of improving translation performance. It does this stuff:

 - Use a 'connection pool' to ensure that translation workers (whether local or bedrock) are always busy - rather than just sending N at a time and waiting for all of the first N to finish before sending another block of jobs
 - Add 'tokens per second' and 'seconds for llm job' metrics so we can monitor performance
 - Add `benchmark-translation` and `llama-server-tunnel` scripts to make it easier to run a test translation (without having to upload something to giant. The benchmark script just takes a port number and runs the translation job against llama-server running on that port. Also add various bits of sample text in russian, french and chinese
 - Switch to using Qwen3-4B (rather than 8B) - for a 10,000 word document this appeared to run 100 seconds faster - 170s vs 270s. The quality for the little bit of french/english translation I eyeballed seemed ok, we'll have to ask some people who speak multiple languages to give some feedback whilst we decide if 4B is 'good enough'.

Prior to deployment, [this](https://eu-west-1.console.aws.amazon.com/systems-manager/parameters/%252FPROD%252Finvestigations%252Ftranscription-service%252Fllamacpp%252FmodelPath/description?region=eu-west-1&tab=Table#list_parameter_filters=Name:%3A:llama) parameter needs updating to point at Qwen3-4b rather than 8B.

Tested fairly extensively using the benchmark script on CODE


twrichards (approved):
🙌

PR #317, opened 2026-08-06
Bumps puppeteer to latest minor to address crtical basic ftp vulnerability

SHession:
## What does this change?

Attempts to address https://github.com/guardian/transcription-service/security/dependabot/147, by bumping puppeteer which is responsible for pulling the vulnerable version of `basic-ftp`.

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #318, opened 2026-08-06
Bumps puppeteer to latest minor to address crtical basic ftp vulnerability

SHession:
## What does this change?

Attempts to address https://github.com/guardian/transcription-service/security/dependabot/147, by bumping puppeteer which is responsible for pulling the vulnerable version of `basic-ftp`.

Closed duplicate #317 due to CI issue with the branch name: https://github.com/guardian/transcription-service/pull/317#issuecomment-5202549366

I'm assuming if CI passes this should be safe.

Vulnerabiliites addressed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3436,15 +3436,17 @@
 			}
 		},
 		"node_modules/@puppeteer/browsers": {
-			"version": "2.10.10",
+			"version": "2.13.2",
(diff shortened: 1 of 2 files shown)


twrichards (approved):
changes LGTM - thanks @SHession 

we tried testing together IRL however there is a more general issue with webpage-snapshot (see https://github.com/guardian/investigations-internal-issues/issues/142) so we should merge this asap to address the vuln and we can resolve any potential issues with this upgrade when addressing the existing timeout issue

PR #321, opened 2026-08-20
change cdk snapshotting to be the actual infra to be created

twrichards:
...rather than `TEST` stage (in doing so capture all of it not just `TranscriptionService`)


philmcmahon (approved):
Nice

